PR #69, opened 2014-12-31
Update hamming_test.spec.js

trayo:
Removing `.compute` from the test makes the user use `module.exports = compute;`, instead of `exports.compute = compute;`, when they create their hamming.js module.

This change keeps this test consistent with the tests prior to hamming.

When I attempted hamming I was getting the error "Undefined is not a function," for the first test, even when I was simply returning 0 (the expected answer). It was difficult to search for an answer on the web and finally found the answer when I compared the hamming_test file with word_count_test file.

### DIFF
--- a/hamming/hamming_test.spec.js
+++ b/hamming/hamming_test.spec.js
@@ -1,4 +1,4 @@
-var compute = require('./hamming').compute;
+var compute = require('./hamming');
 
 describe('Hamming', function () {
 


drueck:
It might be cool to be able to help users out with common errors, especially when something new is first introduced in a given language track. I'm not sure the best way to do that. The README itself is language agnostic (it's the shared across all language tracks that have that same problem) but I think there is the ability to include additional files to be downloaded along side the test file. I suppose one could include a comment in the spec file itself, or include an additional language-specific documentation file such as a `GETTING_STARTED.md` for example as mentioned in the contributing guide.

The only thing I'd be a little bit nervous about would be depriving the users of the ~~pain~~ experience  of carefully examining the test code and trying to figure out what is expected, or of the decision on how to go about making the tests pass. Especially in JavaScript there are a bunch of different ways to do any given thing, so it would perhaps be a shame to suggest one way as if it was the only right way. For example, you could make the tests pass with both of the following structures (as well as many other variations of them):

```javascript
// example.js - assigning an object to module.exports
module.exports = {
  compute: function(strand1, strand2) { ... }
};
```

```javascript
// example.js - assigning a function to exports.compute
exports.compute = function(strand1, strand2) { ... };
```

What do you think? Would there be a way to help people not waste massive time just getting started without giving away too much or guiding them too much to a certain implementation?

You have any thoughts, @kytrinyx, @wilmoore?

wilmoore:
> The only thing I'd be a little bit nervous about would be depriving the users of the pain experience of carefully examining the test code and trying to figure out what is expected

:+1: 

> Would there be a way to help people not waste massive time just getting started without giving away too much or guiding them too much to a certain implementation?

Some pain is good. It leads to more learning. I'd be hesitant if it were a super advanced, but since the answer is easy enough to google , I'd prefer to have people sort of bump up against it and feel the joy of grokking it on their own.

I'm not 100% sure if @kytrinyx will agree with that; so I'm OK with other approaches; though, I think this approach is ideal.

drueck:
Thanks for the feedback, @wilmoore. I agree. While it can be frustrating encountering something like this for the first time without explanation, I think the opportunity for learning it presents is really valuable. You okay with leaving it as is, @trayo?

kytrinyx:
> Some pain is good. It leads to more learning. I'd be hesitant if it were a super advanced, but since the answer is easy enough to google , I'd prefer to have people sort of bump up against it and feel the joy of grokking it on their own.

100 times this.

One of the things that I'm trying to figure out with exercism is how to get people to stop telling other people the answers. There's an amazing feeling that comes with figuring it out on your own.

That said, if the leap is too large, then it just leads to people being stuck, which doesn't lead to learning at all. The balance is the tricky thing.

kytrinyx:
Side note. As a part of issue #67 we should mention how to run the test suite to verify that the tests and examples are consistent.

trayo:
@drueck I definitely agree with struggling through the learning as mentioned by @wilmoore. I think it's great as is then. Thanks!

drueck:
Sounds good. I'll go ahead and close it. Thanks for the PR that resulted in this discussion!

PR #70, opened 2015-01-15
Added single string test to javascript anagram tests

helenst:
I found a submission where the coder was testing for length of arguments to decide what to do... this case wouldn't have worked.

### DIFF
--- a/anagram/anagram_test.spec.js
+++ b/anagram/anagram_test.spec.js
@@ -72,4 +72,10 @@ describe('Anagram', function() {
     expect(matches).toEqual(["tan"]);
   });
 
+  xit("matches() accepts single string argument",function() {
+    var subject = anagram("ant");
+    var matches = subject.matches("tan");
+
+    expect(matches).toEqual(["tan"]);
+  });
 });


kytrinyx:
I'm a bit uneasy about making the test suite define a design where sometimes the function returns one type of object (array) that responds to one type of messages, and at other times returns a different type of object (string) that responds to a completely different set of messages.

I know this is somewhat normal (`undefined` comes to mind in javascript, and `nil` or `null` in other languages), but it's still a design decision that I see causing a lot of pain and complexity.

That said, I think that we should find a way to discourage the behavior you saw. Do you have the entire example or code snippet?

helenst:
Sorry, I'm not sure I was clear enough there - let me try again...

In the exercise the function accepts either a single array of strings, or a number of strings as multiple arguments. I'm just adding a test for the case where it's strings, but only one of them. The return value is expected to be an array in every case, as before.

So as well as testing for:
```
    anagram("asdf").matches(["fdsa", "fish"]) == ["fdsa"]
    anagram("asdf").matches(["fdsa"]) == ["fdsa"]
    anagram("asdf").matches("fdsa", "fish") == ["fdsa"]
```

it now also tests for:
```
    anagram("asdf").matches("fdsa")
```

Here's the submission that promped this PR:
http://exercism.io/submissions/b5b6f121ec0548319da376b768047569 

I hope that made sense.

drueck:
:+1: Thanks, @helenst, seems like a great addition to me as it helps clarify the expected behavior and catches that edge case. With your permission, @kytrinyx, I'd be glad to merge it in.

kytrinyx:
Aaaah, I totally misunderstood. Yepp, that seems good to me. Go for it!

PR #79, opened 2015-04-18
Comment out all triangle tests (except first one)

rootulp:
Made change to conform to the style of previous xjavascript specs. Also
having only one test uncommented suggests incremental development.


kytrinyx:
Thank you

PR #81, opened 2015-04-23
Remove whitespace

rootulp:
Removed trailing whitespace between tests in circular-buffer spec


kytrinyx:
Thank you :heart: It's always nice to get stuff like this tidied up!

PR #87, opened 2015-05-11
Add hello-world exercise

trayo:
I decided to copy the skeleton from the Bob exercise. Will Bob still need it in this case?

Completes Javascript for the hello world list: https://github.com/exercism/rossconf/issues/2

### DIFF
--- a/config.json
+++ b/config.json
@@ -4,6 +4,7 @@
   "repository": "https://github.com/exercism/xjavascript",
   "active": true,
   "problems": [
+    "hello-world",
     "bob",
     "word-count",
     "hamming",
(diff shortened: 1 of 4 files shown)


kytrinyx:
No, Bob probably won't need the skeleton, but we can leave it in for now, I think.

The build is failing:

```
running tests for: hello-world
cp: cannot stat `hello-world/hello-world_test.spec.js': No such file or directory
make[1]: *** [test-assignment] Error 1
make: *** [test] Error 1
```

What do you get if you run `make test` locally?

trayo:
I had the file names wrong. I updated them to match conventions and `make test` passed locally.

kytrinyx:
Excellent, thanks.

PR #88, opened 2015-05-24
[triangle] fixing edgecase of triangle inequality law

schorsch3000:
Hi,
in fact, a triangle with lets say side length 10,10,20 and angles 0,0,180 are quite wired, but are technically triangles.

the law is:
z <= x + y 
not 
z < x + y

### DIFF
--- a/crypto-square/crypto-square_test.spec.js
+++ b/crypto-square/crypto-square_test.spec.js
@@ -16,6 +16,11 @@ describe("Crypto",function() {
     expect(crypto.size()).toEqual(2);
   });
 
+  xit("size of small square with additional non-nuber chars",function() {
+    var crypto = new Crypto('1 2 3 4');
+    expect(crypto.size()).toEqual(2);
+  });
+
   xit("size of slightly larger square",function() {
     var crypto = new Crypto('123456789');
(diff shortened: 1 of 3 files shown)


kytrinyx:
Nice catch, thank you!

The build is failing -- I'm not sure if the example file is outdated, or if the test needs to be tweaked:

```
running tests for: triangle
.............F.
Failures:
  1) Triangle edge cases of triangle inequality are in fact legal
   Message:
     Expected function to throw an exception.
   Stacktrace:
     Error: Expected function to throw an exception.
    at null.<anonymous> (/tmp/triangle.OYNg68BRrG/triangle_test.spec.js:72:42)
Finished in 0.013 seconds
15 tests, 15 assertions, 1 failure, 0 skipped
```

schorsch3000:
Oh i missed to negate the check for this test.
weirdly a make && make test run without errors on my machine...

kytrinyx:
Weird, I wonder what's different.

Would you also mind squashing the commits?

schorsch3000:
There you go.

kytrinyx:
Lovely, thank you!

PR #90, opened 2015-05-26
'octal' is missing a unit test for invalid digits

siebenschlaefer:
There is a unit test, that tests if "carrot" is converted to 0, but there is no unit test for inputs containing invalid digits, e.g. '8'. The example implementation does not validate the input, too.

### DIFF
--- a/octal/example.js
+++ b/octal/example.js
@@ -3,13 +3,18 @@
 var BASE = 8;
 
 function Octal(octal) {
-  this.digits = octal.split('').reverse().map(Number);
+  this.digits = octal.split('').reverse().map(this.digitValue);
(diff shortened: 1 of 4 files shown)


kytrinyx:
Thanks @siebenschlaefer! The pull request has a mix of different changes in it. Would you mind tweaking this to be two different pull requests? Also, if you could ensure that the name of the problem is mentioned in the commit message that would really help, as well (e.g.: `Add missing test case for octal`, or `octal: Add missing test case`).

siebenschlaefer:
OK, did that.

kytrinyx:
Thank you!

PR #92, opened 2015-05-27
octal: added missing test case / fixed implementation

siebenschlaefer:
There is a unit test, that tests if "carrot" is converted to 0, but there is no unit test for inputs containing invalid digits, e.g. '8'. The example implementation does not validate the input, too.


kytrinyx:
Thanks!

PR #93, opened 2015-05-27
queen-attack: added missing test case / fixed implementation

siebenschlaefer:
The unit tests and the example implementation test the north-west / south-east diagonal, but are missing the north-east / south-west diagonal.
I just added two test cases and amended the example implementation.


kytrinyx:
Thank you!

PR #94, opened 2015-06-07
Add test case for leading or trailing whitespace.

eloquence:
Strings with leading or trailing whitespace should not increment the total word count.


kytrinyx:
Nice find, thanks!

PR #96, opened 2015-08-12
Robot name

ZacharyRSmith:


robot-name/robot-name_test.spec.js
```diff
@@ -12,9 +12,17 @@ describe("Robot", function() {
...
   it("different robots have different names", function() {
-    var robotOne = new Robot();
-    var robotTwo = new Robot();
-    expect(robotOne.name).not.toEqual(robotTwo.name);
+    var usedNames= {};
+
+    var i = 0,
+        max = 10000;
+
+    for (i; i < max; i++) {
```
kytrinyx: Is it common to do the variable declaration outside of the loop? In the languages I work in, the preference is usually to declare the variable as close to where it is used as possible. I would also hard-code the max instead of assigning it to a variable here.

kytrinyx:
This looks good, I have the one question about variable assignment outside the loop, but that's it.

robot-name/robot-name_test.spec.js
ZacharyRSmith: Based on what I've seen on SO, eloquentjavascript.net, etc., it is not common. Personally, I also prefer:

    for (var i = 0; i < 10000; i++) {

The reasons I declared outside the loop are:

1) [A style guide](https://github.com/rwaldron/idiomatic.js/) suggests this way. Ctrl+f: "length = 100;"

2) A JSLint tool I ran wanted them declared outside.

3) Apparently it is [the "JSLint" way](http://stackoverflow.com/questions/4646455/jslint-error-move-all-var-declarations-to-the-top-of-the-function), the controversial gold standard for JS. But I agree with the top/accepted answer to the linked SO. Even though declaring outside makes academic sense...blech, I'm abandoning it.

rchavarria: In other languages is common practice to declare a variable as closest as its use as possible. In JavaScript it is recommended to declare them at the beginning of the scope (usually the enclosing `function`, for loops do not create a new scope, as other languages do). JSLint just try to force this rule.

In fact, the recomendation is to declare all variables inside a the same scope in just one `var` statement, see [Move `var` declarations to the top of the function][1].

The reason for this rule is the way JavaScript works with variables (see [*hoisting*][2]). At execution time, JavaScript gets all variables and creates them at the top of the scope. Then, JavaScript assign values where the programmer initializes the variable. This can cause some problems:

```javascript
function output() {
  console.log(foo);
  var foo = 'bar';
  console.log(foo);
}
output();
```

Other languages would throw an error on the first `console.log(foo)`, but the output of this *program* is:

```bash
undefined
bar
```

The way JavaScript works, the previous code is equivalent to this:

```javascript
function output() {
  var foo;
  console.log(foo);
  foo = 'bar';
  console.log(foo);
}
output();
```

All this is just a recommendation, to prevent silly mistakes. So, it would be fine to write the `for` loop in this way:

```javascript
for(var i = 0; i < max; i++) { /* ... */ }
```

as long as you know it won't cause any problem.

[1]: http://jslinterrors.com/move-var-declarations-to-the-top-of-the-function/
[2]: http://www.w3schools.com/js/js_hoisting.asp

kytrinyx: Wow, ok that does make sense. I still like it better with `for (var i = 0; i < 1000; i++)`, to be honest.

I'm not a JavaScript dev, though, so if the two of you can agree on something, I'll leave the final decision to you :)

ZacharyRSmith: I'm not a JS dev, either. While I also like for (var i = 0; i < 1000; i++)
better, I'll defer to rchavarria.

Zachary Smith

rchavarria: Yeah, JSLint rules are sometimes controversial. And you'll find people on both sides of the discussion. I'd write the `for` loop the following way, that is in between the two edge cases we're talking here, declare the variable at the top, initialize it in the loop:

```javascript
it("different robots have different names", function() {
  var i,
    max = 10000,
    usedNames= {};

  for (i = 0; i < max; i++) {
    // ...
  }
}
```

ZacharyRSmith: Commit 8a68287e48b61f7c56ca3e46632c99fed1d803ea follows rchavarria's suggestion.

I'm still new to people caring about how I git. Should I have somehow changed old commits instead of just adding new ones?

kytrinyx: Since this is one atomic change, I think it would make sense to squash all the commits into a single one. Did I link you to this last time? I don't recall: http://blog.steveklabnik.com/posts/2012-11-08-how-to-squash-commits-in-a-github-pull-request

ZacharyRSmith:
Squashed into 1 commit.

Thanks for the link! Btw, I found the "Squash your changes" part of [this](https://github.com/edx/edx-platform/wiki/How-to-Rebase-a-Pull-Request) helpful. It seems to be a good reference for such and rebasing.

kytrinyx:
That's a great link, thanks!

PR #97, opened 2015-08-14
bob-test: COMMENT to clarify umlauts test.

### LINKED ISSUE #77: Clarification regarding bob_test.spec
Guys, is this spec correct?
xit("calmly speaking about umlauts", function() {
    var result = bob.hey("\xfcML\xe4\xdcTS");
    expect(result).toEqual('Whatever.');
  });

Ref: https://github.com/exercism/xjavascript/blob/master/bob/bob_test.spec.js#L61

My understanding says that since the input here has capital letters, they signify yelling at Bob and so the expected result must be "Whoa, chill out!".

ZacharyRSmith:
Fixes #77

### DIFF
--- a/bob/bob_test.spec.js
+++ b/bob/bob_test.spec.js
@@ -54,7 +54,8 @@ describe("Bob", function() {
   });
 
   xit("shouting with umlauts", function() {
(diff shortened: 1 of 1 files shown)


kytrinyx:
Nice!

Issue #91, opened 2015-05-27
Incorrect expectations in "grains" exercise


nonsensery:
In the "grains" exercise, we are asked to write code to calculate the number of grains on the tiles of a chess board where, for the nth tile, the number of grains is 2^(n-1). Since a chess board has 64 squares, that means the final square will have a value of 2^63 (`9223372036854775808`) grains. However, in JavaScript, integers are represented as 64-bit floating point numbers, so they cannot accurately represent this number. For example:

```js
console.log(9223372036854775808);
// =>       9223372036854776000
//                         ^^^^
```

The tests do not take this into account:

```js
  it("square 64", function () {
    expect(grains.square(64)).toEqual(9223372036854775808);
    // => Test is expecting value to equal 9223372036854776000
  });

  it("total", function () {
    expect(grains.total()).toEqual(18446744073709551615);
    // => Test is expecting value to equal 18446744073709552000
  });
```

My guess is that solving this "for real" is outside of the intended scope, since it would require the use of some sort of "big integer" type. Instead, it might make sense to alter the exercise to use a smaller board. The total for a 7x7 board (`562949953421311`) is well within the 2^53 limit for JavaScript integers, but 7x7 is kind of a strange size. Maybe 6x6?

nonsensery:
I suppose another option would be to provide a big integer module in the exercise code. Some options:
- [strint](https://github.com/rauschma/strint)
- [javascript-biginteger](https://github.com/silentmatt/javascript-biginteger)
- [BigInteger.js](https://github.com/peterolson/BigInteger.js)

kytrinyx:
In other language tracks the solutions use BigInteger types. I think it would make sense to include one of these for this problem. Do you have any preferences?

nonsensery:
It seems like [BigInteger.js](https://github.com/peterolson/BigInteger.js) might be a good choice. It has a sane API, and (more importantly) is licensed as public domain.

[Here's a gist with a sample solution](https://gist.github.com/nonsensery/1b48da5316a092dc7312) (I've installed the `big-integer` module using npm).

kytrinyx:
Yeah, I think it would be easier if we included the file. Anything to reduce the friction of having to figure out external dependencies.

ZacharyRSmith:
#98 closes this

kytrinyx:
Thanks, @ZacharyRSmith

PR #98, opened 2015-08-15
grains: Change to use big-integer re: #91.

ZacharyRSmith:
Addresses #91


kytrinyx:
The reference solution for grains is not passing on CI, would you mind taking a look?

ZacharyRSmith:
Sorry I missed that. I'll try to fix it by Monday

ZacharyRSmith:
Hmm...I don't know what I'm doing.

I saw the following in the CI err output:

    The command "make test" exited with 2.
    $ ./bin/fetch-configlet

I also saw fetch-configlet in the .travis.yml file.

So, I looked at the fetch-configlet file. Which led me to [this](https://github.com/exercism/configlet/releases/tag/v1.0.2).

Where I downloaded the source code, which had a README:

    # Configlet

	This is a tool to help verify that an exercism problem repository is valid and up to date.

	There are three common problems that occur:

	1. The `config.json` might be invalid.
	2. Problems might be implemented (test suite + example solution), but not configured.
	3. Slugs in the configuration might not have a corresponding problem.

	## Usage

	```bash
	$ configlet path/to/problem/repository

But ```configlet``` was not a valid command and a quick Google showed no tool to install.

config.json did not help.

I took a quick gander at Travi CI, but found nothing useful.

I looked at a couple of SOs for why their CI tests failed to find modules...upper-vs-lower case, but that is not my problem.

I wondered if it could not resolve file name because of the "-"...so I changed it to "_" but it still failed.

And now I'm tired of puttering around when someone knowledgeable might could fix in 5 minutes. >.<

ZacharyRSmith:
Am reading https://github.com/exercism/x-api/blob/master/CONTRIBUTING.md...

kytrinyx:
Weird, it sounds like the CI test is looking in the wrong place for the biginteger module. Maybe it's looking relative to where the command is being run from, rather than relative to the test file.

ZacharyRSmith:
Copying the big_integer.js file to the xjavascript folder didn't help.

I'm now thinking the Makefile has something to do with it. It has some language specific config. Part of it:

    EXAMPLE := "example.$(FILEEXT)"
    TSTFILE := "$(ASSIGNMENT)_test.spec.$(FILEEXT)"

I think I need to tell it about the big int module somehow.

ZacharyRSmith:
So, I got it passing, but it's hacky. I added big-integer.js to the xjavascript folder, and have the Makefile cp it to the outdir for each "assignment" (exercise) so that it is available for the grains exercise. But it is also being cp'd to the outdir for irrelevant exercises.

I tried to fix this irrelevant cp'ing by changing the Makefile thus: instead of just ```@cp big-integer.$(FILEEXT) $(OUTDIR)```, I tried code along the lines of (pun definitely intended):

    ifeq ($(ASSIGNMENT), grains)
      @cp $(ASSIGNMENT)/big-integer.$(FILEEXT) $(OUTDIR)
    endif

but, depending on what exactly I tried, I got errors such as unexpected end of file, unexpected word, expecting ")", etc. I tried understanding what's going on by SO and the GNU manual, but it seems I'd need a deeper understanding of make and shell to make progress. But I'd prefer spending my time elsewise, and I think this would be a <5min fix for someone who knows what they're doing. So, I'm submitting as-is. I'd definitely understand if you'd prefer not merging this hacky solution. ( :

kytrinyx:
It works, it goes in. Someone will surely be inspired (read: annoyed enough) to fix it at some point :)

kytrinyx:
Thanks!

PR #99, opened 2015-08-17
custom-set: BUG FIX #65. Define empty sets as disjointed.

### LINKED ISSUE #65: Custom Set tests that empty sets are not disjoint.
Custom Set currently tests that empty sets not disjoint and indeed the example program works this way.  This is pretty much at odds with mathematics and with the other test programs and should be changed.

ZacharyRSmith:
Closes #65


kytrinyx:
Thanks!

PR #100, opened 2015-08-17
meetup: CHANGE approach away from metaprogramming RE: #35.

ZacharyRSmith:
Addresses #35

### DIFF
--- a/meetup/example.js
+++ b/meetup/example.js
@@ -1,56 +1,72 @@
-function Meetup(month,year) {
-  'use strict';
-
-  var meetup = {};
-
-  meetup.startingDate = new Date(year,month,1);
-
-  var daysOfTheWeek = [ "Sunday", "Monday", "Tuesday", "Wednesday",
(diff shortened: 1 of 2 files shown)


kytrinyx:
This looks good to me, thanks.

I'm curious why you use upper case for the first word in the commit message. I like commit messages to read like imperative sentences, and the upper case word makes the emphasis sound really weird in my head. What led you to this practice?

ZacharyRSmith:
Thanks for the question/comment. Compared to other people's commits I've been seeing, I make lots of commits. Because I make so many commits, I started upper casing keywords to help me find commits. I think that I'll stop making so many commits (or squashing them) and stop upper casing.

kytrinyx:
I make a ton of commits as well, and then I usually squash them before I submit pull requests.

ZacharyRSmith:
Yeah, I think I'll make a ton of commits, continuing my upper case method for my own benefit, then squash them with a regular sounding message. Thanks for the feedback!

Issue #19, opened 2014-05-20
Revisit ordering of problems


kytrinyx:
In particular, all the song-based/text-based exercises are pretty mentally taxing, and should probably be spread out a bit.

Roshanjossey:
@kytrinyx: Shall we close this issue in favour of issue #299 ?

kytrinyx:
#299 doesn't address the order, but it does put everything in place so that we can address it. I think you're right. Closing.

PR #103, opened 2015-08-20
Re: #19. Change ordering of exercises.

ZacharyRSmith:
Addresses #19 , but I do not think this should close it. I think we should get more meta...maybe even have a spreadsheet or DB that attribute properties to each exercise, so that we could more easily get a bigger-picture view of what the exercise ordering is doing. Also, while I'm dreaming, it would be awesome if we had "paths" so that users could say, "I want to focus on text processing" or "I want to focus on data structures" and have their problems tailored...maybe even have a GUI for this...*drools

I separated the singing exercises, "beer-song" and "food-chain", with a couple of exercises.

I went off of xpython's ordering somewhat, because I've been liking it better than xjavascript's ordering.

nucleotide-count seems a bit much with both hamming and rna-transcription, so I deprecated it (nucleotide-count is also deprecated in xpython).

Brought "leap" and "gigasecond" closer to beginning because they require simple solutions. They are also early in xpython.


kytrinyx:
I think nucleotide-count is what we used to call hamming... not sure how we ended up with both.

These changes are great. I would LOVE to have paths.

With respect to how to get more insight into how things are going, I'm going to start a discussion in the x-common repository and tag you in it. We need to figure out what information would be useful, and then create a mentor/maintainer page that gives us these stats more easily than a spreadsheet!

PR #106, opened 2015-08-30
Improve test specs on "Leap" exercise

rchavarria:
While solving this exercise for exercism.io I thought test specs could be improved (from my point of view, so I might be wrong). This PR tries to clarify the specs and provides some more examples of leap and non-leap years.

Specs haven't changed, but I hope texts describing them are clearer now.

### DIFF
--- a/leap/leap_test.spec.js
+++ b/leap/leap_test.spec.js
@@ -1,19 +1,42 @@
 var isLeapYear = require('./leap');
 
-describe("Year", function() {
-  it("a known leap year", function() {
-    expect(isLeapYear(1996)).toBe(true);
+describe("Leap year", function() {
+
+  it("is not very common", function() {
+    expect(isLeapYear(2015)).not.toBe(true);
   });
 
-  xit("any old year", function() {
-    expect(isLeapYear(1997)).not.toBe(true);
+  xit("is introduced every 4 years to adjust about a day", function() {
+    expect(isLeapYear(2016)).toBe(true);
   });
 
-  xit("turn of the 20th century", function() {
+  xit("is skipped every 100 years to remove an extra day", function() {
     expect(isLeapYear(1900)).not.toBe(true);
   });
 
-  xit("turn of the 21st century", function() {
-    expect(isLeapYear(2400)).toBe(true);
+  xit("is reintroduced every 400 years to adjust another day", function() {
+    expect(isLeapYear(2000)).toBe(true);
   });
+
+  // Feel free to enable the following tests to check some more examples
+  xdescribe("Additional example of a leap year that", function () {
+
+    it("is not a leap year", function () {
+      expect(isLeapYear(1978)).not.toBe(true);
+    });
+
+    it("is a common leap year", function () {
+      expect(isLeapYear(1992)).toBe(true);
+    });
+
+    it("is skipped every 100 years", function () {
+      expect(isLeapYear(2100)).not.toBe(true);
+    });
+
+    it("is reintroduced every 400 years", function () {
+      expect(isLeapYear(2400)).toBe(true);
+    });
+
+  });
+
 });


kytrinyx:
I'm not sure the extra specs help provide any additional information. It seems like they're just more examples of the same rules.

/cc @exercism/track-maintainers What do you think? Does redundancy help make the rules more clear?

hankturowski:
I agree with @kytrinyx. Unless the duplication serves to build illustrate the solution or walks the user in the right direction I don't think it helps.

But... maybe we could remind users that it's okay to add their own unit tests if they want to. The ones we provide aren't carved in stone. The tests @rchavarria added would be good spot checking if you wanted to explore your own solution a bit more.

rchavarria:
I know is better having just a single assertion in each test, but sometimes I find it useful to have more than one example in the same test.

What about this: right now, there are four tests, four specs. I can write each test with just one assertion, as they were before. And I can add a `describe` block with additional examples. The `describe` block can be excluded (it won't be run) or commented. If a user wants to run more test she just needs to activate (remove the `x` in `xdescribe`) or uncomment the block.

verdammelt:
@rchavarria thank you for taking time to try to improve the tests of an exercism track. I espcially like your rewording of the describe blocks. However I must agree with @kytrinyx that I do not think that these extra examples add much. 

At most, a single extra example of each part of the rule *may* be useful. Each new test/assertion adds a maintenance cost so we should keep as few tests as possible (but no fewer!). Do the additional examples show interesting corner cases of the Gregorian Calendar leap year rule?

(excuse me while I put on my pedantic hat...)

Your additional examples specify that the function should compute leap-year-ness in the Proleptic Gregorian Calendar (https://en.wikipedia.org/wiki/Gregorian_calendar#Proleptic_Gregorian_calendar). Was this intentional? Any date before 1582-10-15 is a HUGE corner case of the Gregorian Calendar. And things are complicated by geography: because I am located in the USA (a former British Colony) any date before 1752-09-14 is the limits of the calendar; other localities changed to the Gregorian Calendar at different times (https://en.wikipedia.org/wiki/Gregorian_calendar#Adoption).

I think that this exercise should quietly steer well clear of the common problematic dates (1582 & 1752) by sticking to dates after 1800 (although those also may fall into corner cases depending upon locale). 

As an 'extra credit' I have prompted one or two submitters on the xlisp track to add tests & code to handle dates before the time of the Papal Bull of 1582.

(takes off pendantic hat...)

Programming with Time is hard, this exercise seems to be meant to be about integer division and modularization of the code - let's keep the exercise simple, it only needs a few examples.

rchavarria:
Ooops. My apologies. It wasn't my intention to test such corner cases. Yep, I'm aware of some *interruptions* or particularities in the Gregorian calendar, but I didn't take them into account here. I was thinking of years just as integers, as @verdammelt said: test should mean just integer division and modularization. I understand that using years below 1800 can lead to confusion.

Wow, I didn't think adding a couple of examples would create such a great discussion. Thanks everybody.

I will write a main `describe` block, with just one assertion for each existing test case. And an additional `xdescribe` block with another example for each test case. I'll take care of corner cases due to the Gregorian calendar (maybe, just using future years).

kotp:
Good explanation, including the regional issues.  There is room here as well to not simply do integer division, as you can express a year as a float (and this is commonly done, but usually behind the scenes).  But it is appropriate to avoid it.

As for additional assertions in the tests, I agree that it is sometimes helpful, but they should be border examples in the category of the test.  If they can communicate clearly, without being strictly redundant, then they may be beneficial.

Otherwise they are 'fluff'.

rchavarria:
PR updated with changes to have just one assertion for each test and an additional `xdescribe` block to let users enable it if they want to check more examples.

kytrinyx:
I'm happy with this version. Anyone else have any additional comments before we merge this?

kotp:
:+1:  from me

hankturowski:
Looks great.

kytrinyx:
Thanks, everyone for the great discussion. Thanks @rchavarria for working through this!

rchavarria:
My pleasure, thanks everybody for helping and teaching me.

PR #107, opened 2015-09-13
hamming: test requiring a function opposed to an object

Moezalez:
For more consistency. 
Most tests (at least the first few) require a function opposed to an object. There is no reason to break with this habit in this particular test.


kytrinyx:
Yeah, hamming doesn't need an object. Thanks! Would you mind squashing the commits, please?

Moezalez:
Done :)

kytrinyx:
Thanks!

PR #108, opened 2015-09-14
Removing inconsistent use of quotes and adding semicolons.

Moezalez:
I noticed that some of the time double quotes and some of the time single quotes were used.
I changed all double quotes to single quotes to remove this inconsistency via ESLINT. 

Also I added missing semicolons via ESLINT.


kytrinyx:
Nice, the consistency is a definite improvement. Thanks!

PR #110, opened 2015-09-18
Fix spuriously passing tests.

wobh:
This should not pass:

```javascript
module.exports = function (year) {
  return "Foo"
};
```

### DIFF
--- a/leap/leap_test.spec.js
+++ b/leap/leap_test.spec.js
@@ -3,7 +3,7 @@ var isLeapYear = require('./leap');
 describe('Leap year', function() {
 
   it('is not very common', function() {
-    expect(isLeapYear(2015)).not.toBe(true);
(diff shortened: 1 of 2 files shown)


rchavarria:
Your proposed code doesn't make all tests to pass, so it wouldn't be a valid solution to the exercise. But I see the point of your changes and I agree with you. Now they are better tests.

Thanks!

wobh:
I found corresponding `not.toBe` tests in "pythagorean-triplets" and "grains". A fix for the one in "pythagorean-triplets" has been added to this PR, since it was straightforward. In "grains" the tests are that the expectation is "not.toBe" some non-boolean values. That seems dubious to me, but I'm not versed in Jasmine or JS to say what a better alternative would be.

PR #112, opened 2015-10-02
JS crypto_square: delete 2 tests

ZacharyRSmith:
The README has nothing about `normalizeCiphertext()`. Also, the first of these tests makes no sense...where did the expect "s" come from?


kytrinyx:
You're right. I think the fix should be to change the test suite so it expects a square (newlines instead of spaces in what was the normalizedCipherText).

matthewmorgan:
Merged with commit a02ce8b.  Thanks!

PR #114, opened 2015-10-03
Ignore all tests but first

rchavarria:
A very simple PR, mark all tests but first as ignored. 

It surprised me when I solved this problem for exercism. I had lots of tests failing at the beginning and it was because all tests were *active*.


kytrinyx:
Good catch! Thanks :)

PR #115, opened 2015-10-04
Add test cases for rna to dna transcription to enrich example

MarcosX:
Saw that in the ruby example it consider conversion in both ways, which is not exactly the same and might make programmers consider how to remove the duplicated code.
Hope it helps!


kytrinyx:
Thanks!

PR #116, opened 2015-10-06
Fix Hello World Tests

murielsilveira:


kytrinyx:
We've made the second test pending on purpose. In this case it doesn't make a huge difference, since there are only two tests, but in the later exercises, there are many tests, and this avoids having an overwhelming amount of failing tests all at once. It simulates a test-driven workflow.

Since the hello world exercise is an introduction to how the javascript exercises are put together, we felt that it's important to make it so that the programmer must make the pending test active by editing the file.

I'm going to close this, but if you have other arguments that we've not thought through, please do discuss! We're always willing to re-evaluate how we do things.

PR #118, opened 2015-10-17
bracket-push: Add test for proper nesting

petertseng:
This rules out solutions that simply count the number of open/close
brackets without regard for whether they are properly nested.

I submitted a solution (in another language track) that would have
failed this test and did not realize it until looking at others'
solutions. It would be great if I could have known before submitting,
thus I add this test.

Corresponding PR on xgo https://github.com/exercism/xgo/pull/176

### DIFF
--- a/bracket-push/bracket_push_test.spec.js
+++ b/bracket-push/bracket_push_test.spec.js
@@ -21,6 +21,10 @@ describe('bracket push', function() {
     expect(bracket('{[]}')).toEqual(true);
   });
 
+  xit('rejects brackets that are properly balanced but improperly nested', function() {
+    expect(bracket('{[}]')).toEqual(false);
+  });
+
   xit('checks bracket closure with deeper nesting', function() {
     expect(bracket('{[)][]}')).toEqual(false);
   });


bracket-push/bracket_push_test.spec.js
```diff
@@ -21,6 +21,10 @@ describe('bracket push', function() {
     expect(bracket('{[]}')).toEqual(true);
   });
 
+  xit('rejects brackets that are properly balanced but improperly nested', function() {
+    expect(bracket('{[}]')).toEqual(false);
```
petertseng: Reasoning for this being false: https://github.com/exercism/x-common/blob/master/bracket-push.md says **nested correctly**. Please sanity check on whether you agree with this interpretation.

If there is disagreement, we should at least make sure we reach the same conclusions as xgo (see linked PR for that)

rchavarria:
It's nice to have this new test. As you said, it make fail those solutions that only take into account the number of brackets involved and not their positions.

Thank you!

bracket-push/bracket_push_test.spec.js
kytrinyx: I completely agree. This is great, thanks!

PR #95, opened 2015-08-08
adding binary-search problem

NathanEpstein:
Adding the binary search problem to the JS track. Please let me know if there are any changes required!

PR #121, opened 2015-10-22
Extra tests for the 'binary' exercise.

creature:
_cf. [x-common issue #95](https://github.com/exercism/x-common/issues/95)._

I've expanded the JS test for the 'binary' exercise to include the same examples as the [Ruby equivalent](https://github.com/exercism/xruby/blob/master/binary/binary_test.rb). The example solution has also been updated so it passes the new tests.

However, I'm not familiar enough with the Exercism progression to know if this is pitched at an appropriate level – that is, if students are expected to do string validation or understand the nuances of `parseInt` at this point. [JavaScript's `parseInt`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseInt) is rather more flexible than solutions in other languages, and doesn't make it easy to bail out on invalid input: 

> If `parseInt` encounters a character that is not a numeral in the specified radix, it ignores it and all succeeding characters and returns the integer value parsed up to that point... If the first character cannot be converted to a number, `parseInt` returns `NaN`.

If there's a different approach you'd like me to take with this, let me know and I'll adjust this PR to suit. :)

### DIFF
--- a/binary/binary.spec.js
+++ b/binary/binary.spec.js
@@ -1,8 +1,11 @@
 var Binary = require('./binary');
 
 describe('binary', function() {
+  it('0 is decimal 0', function() {
+    expect(new Binary('0').toDecimal()).toEqual(0);
+  });
 
-  it('1 is decimal 1', function() {
+  xit('1 is decimal 1', function() {
     expect(new Binary('1').toDecimal()).toEqual(1);
   });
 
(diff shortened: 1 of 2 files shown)


kytrinyx:
This looks good. I think the progression is still fine the way it is. @rchavarria do you want to have a look at this and merge if you're happy with it?

rchavarria:
It looks ok. In JavaScript, `parseInt` parses `"10nope"` as a valid number, but `README` file asks for handling invalid outputs. Even though `parseInt` returns a number, I agree it is an invalid output in this case.

I would merge this PR but it has conflicts. I think is because of file names. `binary-test.spec.js` file is now called `binary.spec.js`. Would you mind to resolve conflicts @creature? I don't know if it would be better to amend this PR or create a new one, it's up to you.

kytrinyx:
The conflict should resolve itself fairly easily if you rebase onto the most recent upstream master.

kytrinyx:
Also: you can replace a pull request by doing a `git push -f` to the branch.

creature:
@rchavarria & @kytrinyx: I'm glad to hear that it's at an appropriate level. I've rebased and pushed, so it should merge cleanly now. I had to do that last week too, when underscores became dashes, so it wasn't a problem. :)

kytrinyx:
Sweet! Thanks so much.

rchavarria:
Thanks @creature

PR #122, opened 2015-10-23
circular-buffer: Test forced writes on non-full buffer

petertseng:
Other language tracks have been adding this:

Ruby:
https://github.com/exercism/xruby/commit/5560b7a0e032bdf4bca753d4aba4625340853d1c

Go:
https://github.com/exercism/xgo/commit/6b41d2f925a7b35140701a8290e7d9a6155ac315

Rust:
https://github.com/exercism/xrust/commit/e4990f1b0e8c9d16053012b13aaaeb03ac141383

Seems prudent for Javascript to test it as well. This will see if any
solutions are unconditionally dropping when overwriting (that would drop
elements from non-full buffers)

### DIFF
--- a/circular-buffer/circular-buffer.spec.js
+++ b/circular-buffer/circular-buffer.spec.js
@@ -82,6 +82,15 @@ describe('CircularBuffer', function() {
     expect(buffer.read).toThrow(bufferEmptyException());
   });
 
+  xit('forced writes act like write in a non-full buffer', function() {
+    var buffer = circularBuffer(2);
+    buffer.write('1');
+    buffer.forceWrite('2');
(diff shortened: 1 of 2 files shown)


petertseng:
Aaaand I had to edit the example solution as well, otherwise it fails the newly-added test.

kytrinyx:
Sweet. Would you mind rebasing onto master, and squashing the two commits, since they're really about the same change?

petertseng:
sure thing, rebased and squashed

kytrinyx:
I think this seems fine. The overarching goal of the change was to fix a bug/problem with the exercise, not the specifics of how that had to be done (adding a test, changing the solution).

PR #123, opened 2015-10-23
Removed "test" from spec filenames

AbdullahAlger:
In order to make the spec filenames consistent across the xjavascript and xecmascript repositories, I've removed the word "test" from the filenames.


AbdullahAlger:
Sorry, the first commit didn't include the change in Makefile, which made the build not pass. I made another commit and squashed it into the first and the build passed.

PR #124, opened 2015-10-25
Ignore all tests but first

rchavarria:
This is not a great contribution, but I like when all tests but first are ignored.


kytrinyx:
This is a perfectly good contribution :) Thanks

PR #125, opened 2015-10-30
Test immutability of the gigasecond

TFarla:
I have seen submissions which mutate the original date and return that. This works for the current test suite. However, when the `date` method gets called multiple times. That solution adds another gigasecond on a date where one is already applied.

For example. One might implement the date function like this:

```
Gigasecond.prototype.date = function() {
  b = new Date(this.birthday.setSeconds(this.birthday.getSeconds() + 1000000000 ));
  return new Date(b.getFullYear(), b.getMonth(), b.getDate());
};
```

This is the incorrect behaviour that the new test case will catch:
```
var d = new Gigasecond(new Date(1995, 8, 27))
d.date()
// Sat Jun 05 2027 00:00:00 GMT+0200 (CEST)
d.date()
// Tue Feb 11 2059 00:00:00 GMT+0100 (CET)
```

### DIFF
--- a/gigasecond/gigasecond.spec.js
+++ b/gigasecond/gigasecond.spec.js
@@ -20,4 +20,10 @@ describe('Gigasecond', function() {
     expect(gs.date()).toEqual(expectedDate);
   });
 
+  xit('test 4', function() {
+    var gs = new Gigasecond(new Date(1959, 6, 19));
+    var expectedDate = new Date(1991, 2, 27);
+    gs.date();
+    expect(gs.date()).toEqual(expectedDate);
+  });
 });


kytrinyx:
Thanks! I've seen this mutability/immutability problem in some solutions in the Ruby track as well. I'm leaving this for the JavaScript track maintainers to consider.

rchavarria:
I like the idea of immutability. The new test is welcomed.

I would ask for a better name for the test, but it's a bit unfair asking for that when other tests don't have very good names.

kytrinyx:
Yeah, we could write up an issue to suggest that the test names be improved, that way whoever is inspired to work on that could do so in a separate PR.

@rchavarria @mixolidia merge at will.

Issue #128, opened 2015-11-12
All exercises should require a homogeneous structure to be exported


rchavarria:
Take a look at [exercism/exercism.io issue #2304](https://github.com/exercism/exercism.io/issues/2304). It talks about requiring the user to export a homogeneous structure for all exercises.

Some require the user to export a constructor function:

```
var Bob = function() {};
Bob.prototype.hey = function(input) { ... }
module.exports = Bob;
```

Some require to export an object:

```
var Hamming = {
    compute: function(A, B) { ... }
}
module.exports = Hamming;
```

Some require to export different things.

Could we first agree on what structure should exercises require to export?

matthewmorgan:
I agree there should be more consistency, although I think it's worth
pointing out that wrestling with the output formats did help me sharpen my
understanding of modules.

On Thu, Nov 12, 2015 at 11:32 AM, Rubén Chavarría <notifications@github.com>
wrote:

> Take a look at exercism/exercism.io issue #2304
> <https://github.com/exercism/exercism.io/issues/2304>. It talks about
> requiring the user to export a homogeneous structure for all exercises.
>
> Some require the user to export a constructor function:
>
> var Bob = function() {};
> Bob.prototype.hey = function(input) { ... }
> module.exports = Bob;
>
> Some require to export an object:
>
> var Hamming = {
>     compute: function(A, B) { ... }
> }
> module.exports = Hamming;
>
> Some require to export different things.
>
> Could we first agree on what structure should exercises require to export?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/exercism/xjavascript/issues/128>.
>



-- 
Matt Morgan
Director of Technology
Dulcet Software LLC
(734) 926-8142

kytrinyx:
How about making the format consistent for the first 10 or 12 exercises, and then letting things be more inconsistent after people have gotten a few successes?

matthewmorgan:
That sounds good.  That way they'll have some confidence in the internals
of their solution.

On Thu, Nov 12, 2015 at 5:21 PM, Katrina Owen <notifications@github.com>
wrote:

> How about making the format consistent for the first 10 or 12 exercises,
> and then letting things be more inconsistent after people have gotten a few
> successes?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/exercism/xjavascript/issues/128#issuecomment-156255014>
> .
>



-- 
Matt Morgan
Director of Technology
Dulcet Software LLC
[...]

PR #130, opened 2015-11-24
Make tests require the solution to export a classy function

rchavarria:
As commented on #128, the first exercises require a similar structure to be exported on user's solutions.

### DIFF
--- a/anagram/anagram.spec.js
+++ b/anagram/anagram.spec.js
@@ -1,79 +1,79 @@
-var anagram = require('./anagram');
+var Anagram = require('./anagram');
 
 describe('Anagram', function() {
 
   it('no matches',function() {
-    var subject = anagram('diaper');
+    var subject = new Anagram('diaper');
     var matches = subject.matches([ 'hello', 'world', 'zombies', 'pants']);
 
     expect(matches).toEqual([]);
   });
 
   xit('detects simple anagram',function() {
-    var subject = anagram('ant');
+    var subject = new Anagram('ant');
     var matches = subject.matches(['tan', 'stand', 'at']);
 
     expect(matches).toEqual(['tan']);
   });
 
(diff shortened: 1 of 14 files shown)


matthewmorgan:
Looks good.  

However, I noticed something odd when running ```make test```.  It looked like only the un-x-ed out tests are being run by the makefile.

The other thing I noticed is that the tests for ```two-bucket``` are all enabled, so all four tests ran.

I am going to merge this, but it seems like we should create new issues for the makefile and correcting ```two-bucket.spec.js``` so that only the first test is enabled by default.

@rchavarria -- what do you think?

matthewmorgan:
@rchavarria BTW thanks-- this must have been a lot of work!

rchavarria:
@matthewmorgan Totally agree with fixing `two-bucket` and make just the first test enabled.

About `make test` issue -- that's strange. I modified nothing about how tests run. And I noticed nothing. It seems they run ok on Travis CI: https://travis-ci.org/exercism/xjavascript/builds/93075730. I see more than one test on each exercise.

matthewmorgan:
I looked into it and it seems it's a difference in the [OSX implementation](http://stackoverflow.com/questions/2320564/variations-of-sed-between-osx-and-gnu-linux) of ```sed -i``` vs the Linux one.

The ES track uses a different script than the JS track for modifying the test files before evaluation.  This line can be altered to fix it on OSX.

```
-   @sed -i.original 's/\bxit\b/it/g' $(OUTDIR)/*spec.$(FILEEXT)
+   @sed 's/xit/it/g' $(ASSIGNMENT)/$(TSTFILE) > $(INTDIR)/$(TSTFILE)
```
I will put in a PR for this.  Also, "meetup" has all the tests enabled by default, so I will put in a PR for that too.

PR #132, opened 2015-11-25
Modify Makefile sed command for OSX

matthewmorgan:
OSX implements ```sed -i``` differently than Linux, which was causing the makefile to skip any tests marked ```xit``` rather than ```it```.


kytrinyx:
I've run into this particular thing before, and it catches me by surprise every single time!

PR #133, opened 2015-11-25
Disable all tests but first by default in meetup assignment

matthewmorgan:
Meetup assignment had all tests enabled by default.  This fixes that.


rchavarria:
Thanks! 

(Again, I don't know why it appears as closed. I thought I got it :disappointed:)

PR #134, opened 2015-11-25
Disable two bucket tests

### LINKED ISSUE #131: Disable all tests but first in Two Bucket exercise
Each exercise contains several tests (several `it` Jasmine blocks). To help users to implement the solution, all tests but first should be disabled (they should be `xit` instead of `it`).

matthewmorgan:
Closes #131


rchavarria:
Thanks man!

PR #135, opened 2015-11-25
Clean up docs for x-api v3

kytrinyx:
We're rewriting the language pages on the site to include the documentation, which can now be served via the API.

It's much cleaner--and it means that the documentation is in the repository for the exercises that it references, which is convenient, since the track-maintainers are the ones who should have access to reviewing and merging it.

This is a direct port of the documentation from the docs repository.


rchavarria:
Some files are deleted while other are mixed together in a different file. Is that ok? Do all language track have the same `docs` folder structure?

kytrinyx:
The old docs weren't used anywhere, so that's fine. All the languages have the same file structure. We may add some more files (e.g. `HINTS.md`), but for now we are operating with only the following:

- about
- installation
- tests
- resources
- learning

Where resources are generically useful things, and learning is specifically resources that would help people learn the language from scratch.

rchavarria:
Ok, nice to know! Thanks!

I'll let @matthewmorgan to merge it.

matthewmorgan:
@kytrinyx thanks!

PR #136, opened 2015-11-26
Add new exercise pangram

matthewmorgan:
As proposed in exercism/x-common#133


rchavarria:
Thanks for this new exercise!

PR #141, opened 2015-12-01
Add test to help implement Robot with smaller steps

rchavarria:
Same as [exercism/xecmascript PR#62](https://github.com/exercism/xecmascript/pull/62):

> This new test tries to guide the user with smaller steps. One can first hardcode the first name. Then, the user needs to generate a different name, but only one more. After that, the next test ask the user to generate 10k different names.

### DIFF
--- a/robot-name/robot-name.spec.js
+++ b/robot-name/robot-name.spec.js
@@ -22,6 +22,11 @@ describe('Robot', function() {
   });
 
   xit('different robots have different names', function() {
+    var differentRobot = new Robot();
+    expect(differentRobot.name).not.toEqual(robot.name);
+  });
+
+  xit('there can be lots of robots with different names each', function() {
     var i,
         numRobots = 10000,
         usedNames = {};


matthewmorgan:
@rchavarria some of the test suites are specified in exercism/xcommon, but it looks like there isn't an entry for robot-name.  Which means there is no specified common test suite for this exercise, right?

Anyway, I think it's a good idea, and in keeping with TDD.  Thanks for doing it.

rchavarria:
Yep. In xcommon you'll find common specifications, language agnostic, so that each language track can implement a *test writer* that takes the specification and generates a file with tests specific for that language track. 

If there is no entry is x-common is because nobody has created it for Robot name.

Implement this *test writers* or adding more common specs to x-common could be an interesting task. I don't know too much about it, but [x-common/README](https://github.com/exercism/x-common/blob/master/README.md) file could be a good starting point, and [x-common/docs](https://github.com/exercism/x-common/tree/master/docs) folder too.

kytrinyx:
https://github.com/exercism/todo/issues/138 :)

rchavarria:
Much cleaner, thanks!

PR #142, opened 2015-12-02
rna-transcription: Corrected test descriptions

jdfm:
Some of the tests were describing the 'toRna' features when they should have been describing the 'toDna' features.


kytrinyx:
Good catch.

On a slightly related note: We've had a bit of a mix-up with this problem... there's no such thing as transcribing RNA to DNA (biologically speaking). For now let's run with this, I'll make a note to open up some tickets to fix this across the board for all the tracks that implement this exercise.

rchavarria:
Whoops. Copy/Paste side effects. 

Nice catch @jdfm

PR #146, opened 2015-12-29
Update allergies.spec.js

AndreaCatalucci:
I believe the test is wrong: peanuts is missing from the list of expected allergies


matthewmorgan:
@AndreaCatalucci I take it since you closed this you figured out that the test is OK as is?

AndreaCatalucci:
Thanks for your fast reply! Yes I figured that out, but I feel that the readme should be more specific. I will submit a new pull request with my proposal soon.

PR #147, opened 2015-12-30
Update setup.md

znja:
Updated the URL from `http://help.exercism.io/getting-started-with-javascript.html` to `http://exercism.io/languages/javascript` since http://help.exercism.io/ has been deprecated.


matthewmorgan:
Thanks @znja!

PR #148, opened 2015-12-30
Removed space on line 15.

oalbe:
It wasn't supposed to be there and made the whole test fail.


matthewmorgan:
So simple, but so problematic!  Thanks @oalbe !

oalbe:
Anytime.

Issue #154, opened 2016-01-25
Create test pattern in `config.json` to help exercism.io to identify files that contain tests


rchavarria:
There is a small bug in exercism.io ([exercism.io#2610](https://github.com/exercism/exercism.io/issues/2610)). JavaScript tests files are not shown.

It was fixed for ECMAScript track by adding a `test_pattern` property in `config.json` file (see [xecmascript#65](https://github.com/exercism/xecmascript/pull/65/files)).

kytrinyx:
+1, I just looked at this, and confirmed that adding the test_pattern property in the config will fix it.

matthewmorgan:
Great, when should it deploy?  I checked just now and no test suite is visible in the GUI.

kytrinyx:
I am deploying now.

kytrinyx:
I've confirmed the fix.

matthewmorgan:
😃

PR #155, opened 2016-01-25
Modify config.json to add test pattern

matthewmorgan:
As described in #154.

@rchavarria is this what we needed?


rchavarria:
Yep! I think so. We'll see when it's merged and integrated into exercism.io

PR #157, opened 2016-02-09
Add test to catch non-numeric sorting

mgmatola:
A common solution to this exercise involves first sorting the sides of the triangle. This makes both validating input and determining the kind of triangle easier. However, many solutions erroneously sort the sides alphabetically (the default sort for arrays in JS is alphabetical, even if the array elements are numeric). There is already a test case (Triangle(10,10,2)) that correctly fails when determining kind with incorrectly sorted sides. The test case this PR introduces will also fail validation (triangle inequality) with incorrectly sorted sides.

### DIFF
--- a/exercises/triangle/triangle.spec.js
+++ b/exercises/triangle/triangle.spec.js
@@ -77,4 +77,9 @@ describe('Triangle', function() {
     expect(triangle.kind.bind(triangle)).toThrow();
   });
 
+  xit('triangles violating triangle inequality are illegal 3', function() {
+    var triangle = new Triangle(10,1,3);
+    expect(triangle.kind.bind(triangle)).toThrow();
+  });
+
 });


matthewmorgan:
@rchavarria I discussed this PR with @mgmatola and think it is a good addition.  This default sorting behavior is a quirk of JavaScript, and I'm not sure it belongs on every language track, but seems like a good fit for JS and ES too.  What do you think?

rchavarria:
Yep, I've read you conversation on exercism.io. And, if JavaScript has some particular behaviour while sorting arrays, it belongs to the language, so I don't find any problem to consider it although other languages don't.

Feel free to merge it, thanks for the contribution and your comments @mgmatola!

matthewmorgan:
Thanks @mgmatola !

@rchavarria I will add this test to the ES track

rchavarria:
Perfect!

2016-02-09 22:41 GMT+01:00 Matthew Morgan <notifications@github.com>:

> Thanks @mgmatola <https://github.com/mgmatola> !
>
> @rchavarria <https://github.com/rchavarria> I will add this test to the
> ES track
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/exercism/xjavascript/pull/157#issuecomment-182089312>.
>



-- 

Rubén Chavarría
  blog <http://rchavarria.github.io>    linkedin
<http://es.linkedin.com/in/rchavarriat>

<http://www.deliriumboutique.com/>

PR #158, opened 2016-02-12
No tabs

ThomasZumsteg:
Replaced tabs with spaces in 
two-bucket/
 - two-bucket_test.spec.js
 - example.js
 - example-2.js
crypto-square/
 - example.js

### DIFF
--- a/crypto-square/example.js
+++ b/crypto-square/example.js
@@ -1,55 +1,55 @@
 'use strict';
 
 module.exports = function(input) {
-	this.input = input;
+  this.input = input;
 
-	this.normalizePlaintext = function() {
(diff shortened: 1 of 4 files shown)


matthewmorgan:
@ThomasZumsteg thanks for taking this on.

A few things will need to be changed before this can be merged.  

1. It looks like you might have been working with an outdated copy of the master repo.  For example, the tests have all been renamed to `*.spec.js`, rather than `*_spec.js`.  If you need help syncing your fork just ask.

2. The exercise should only have one `example.js`, so `example-2.js` should be removed.

3. Finally, we like to see all changes squashed in to one commit if possible.

I am happy to help you with any of the above steps if that would make it easier.  If you have any questions or need anything at all, just let me know.  Thanks!

PR #160, opened 2016-02-13
Fix markdown syntax in ABOUT.md

suzannehamilton:
This fixes a layout bug on http://exercism.io/languages/javascript.

I've also tidied up the punctuation and trailing whitespace.

The Redcarpet Markdown parser requires a blank line before lists ([see this Babelmark](http://johnmacfarlane.net/babelmark2/?text=You+should+learn+JavaScript+because%3A-%0A*+It's+easy+to+learn.%0A*+It's+versatile+in+the+sense+that+it's+multi-paradigm+supporting+procedural%2C+event+based%2C+object+oriented+and+functional+programming.%0A*+It's+Open+Source.%0A*+JavaScript+programming+skills+are+in+high+demand.)), so the lists in this file were being rendered inline rather than as a `<ul>`.

The bug isn't obvious when you view the file on Github because Github's Markdown parser is more flexible about blank lines.

As a side-note, Redcarpet has a `:lax_spacing` option which you could use if this is a common problem and you don't mind the Markdown parsing being less strict.


matthewmorgan:
Thanks @suzannehamilton !!

PR #163, opened 2016-02-21
Adding a missing isosceles case

mikegee:
pointed out by @plinto in http://exercism.io/submissions/1a614864d2dc44bdb18e1be9f5f6970e

### DIFF
--- a/exercises/triangle/triangle.spec.js
+++ b/exercises/triangle/triangle.spec.js
@@ -17,6 +17,11 @@ describe('Triangle', function() {
     expect(triangle.kind()).toEqual('isosceles');
   });
 
+  xit('isosceles triangles have first two sides equal', function() {
+    var triangle = new Triangle(2,2,3);
+    expect(triangle.kind()).toEqual('isosceles');
+  });
+
   xit('isosceles trianges have first and last sides equal', function() {
     var triangle = new Triangle(4,3,4);
     expect(triangle.kind()).toEqual('isosceles');


rchavarria:
Thank you for the new test case, but I'm sorry, I think that test case is already covered. In fact, I realized we already have two similar cases (third with your proposed one).

You propose: `new Triangle(2, 2, 3)` should be `isosceles`.

On line 30 in your commit, we already have: `new Triangle(4, 4, 3)` expecting it to be `isosceles`.

On line 35, we also have: `new Triangle(10, 10, 2)` expecting it to be `isosceles` as well.

I'm probably missing something, but they look to me identical. @matthewmorgan has been involved in previous modifications of this exercise. Do you have some comments?

mikegee:
I believe the existing tests are incomplete because they include only acute isosceles triangles and no obtuse isosceles triangles. If the tests were more complete, I would have noticed the bug in my submission.

matthewmorgan:
@mikegee I see what you're saying.  There's no overall spec for Triangle in x-common, but the neither Ruby and Java track have tests for obtuse isosceles triangles.

I don't have any objection, @rchavarria ?  Maybe it would make sense to add this test to those tracks too.

rchavarria:
In my opinion, the 3 triangles...:

1. `new Triangle(2, 2, 3)`
2. `new Triangle(4, 4, 3)`
3. `new Triangle(10, 10, 2)`

...test the same idea. An isosceles triangle must have two sides equal, and a different third side. The exercise says nothing about acute or obstuse isosceles triangles. For me it doesn't matter if the triangle is acute or obstuse, it's just isosceles. And tests already cover the three cases for an isosceles triangle: `[3, 4, 4]`, `[4, 3, 4]` and `[4, 4, 3]`.

Sorry, but I don't see the value of adding a new test case that doesn't test anything different. If tests checked for acute/obstuse triangles, I would totally agree. But it's not the case, sorry.

matthewmorgan:
@rchavarria So you feel it's OK if an iteration that doesn't check for obtuse isosceles can still pass the tests?

I see both sides-- @mikegee 's solution was incomplete because it didn't test all three sides, but the tests also allow it to pass.

rchavarria:
Whoops! My fault, sorry. I knew I was missing something.

Of course @matthewmorgan I'm not ok with a wrong solution. I didn't see that `sides` variable in @plinto's solution was sorted. I tested @plinto's code and it fails with @mikegee's test.

My apologies @mikegee, I didn't understand completely the issue :cry: 

I'm looking forward to see this PR merged. And, I'm ok with creating a issue in x-common to discuss with other track maintainers if they think it would be necessary to add this spec to the exercise.

Sorry @mikegee for not taking the time to fully understand you, thanks @matthewmorgan for opening my eyes.

matthewmorgan:
Thanks @mikegee @rchavarria !

PR #164, opened 2016-02-23
Remove test which allows degenerate (a+b=c) triangles

matthewmorgan:
I'm removing this test that i added after the outcome of [a discussion on the Erlang track](https://github.com/exercism/xerlang/issues/85) that concluded we should not allow 'degenerate' triangles, EG sides a,c,b where a+b=c.

### DIFF
--- a/exercises/triangle/triangle.spec.js
+++ b/exercises/triangle/triangle.spec.js
@@ -72,11 +72,6 @@ describe('Triangle', function() {
     expect(triangle.kind.bind(triangle)).toThrow();
   });
 
-  xit('edge cases of triangle inequality are in fact legal', function() {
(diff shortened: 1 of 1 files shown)


rchavarria:
Yeah!

I'll merge this as soon as I have some time. Feel free to merge yourself if you want @matthewmorgan.

What about xecmascript? Is there the same case there?

matthewmorgan:
Yes there is, bc yours truly made the same change on that track!  I'm about to push a fix, then I thought I'd add @mikegee's new test in a separate PR, OK?

matthewmorgan:
For some reason the Travis build isn't starting to run...

rchavarria:
Thanks so much for this!

PR #166, opened 2016-02-25
Update node version

matthewmorgan:
NodeJS is now on version 5.7, so it seems like the thing to do!


rchavarria:
Yep, you're right. I'm sure new users are using newer versions than `0.10` :wink:

PR #167, opened 2016-03-02
Removed test which contradicted the README.md

cebial:
The README.md states:
> To keep your implementation simple, the tests will not cover error conditions. 
> Specifically: pop or shift will never be called on an empty Deque

I removed the one test which did a pop() on an empty Deque anyway.


matthewmorgan:
Well-spotted @cebial!  Thank you.

PR #168, opened 2016-03-07
Use containerized build on travis

kytrinyx:
We don't need to use sudo, and this should (technically) be faster, overall.


matthewmorgan:
Thanks!

PR #172, opened 2016-03-22
make hex characters easier to understand

YanKozlovskiy:
This update will fix the example on the hex characters by telling the reader what each hex character is in normal text format.


matthewmorgan:
Looks good-- thank you!

PR #174, opened 2016-03-25
Remove default multiples

matthewmorgan:
Addresses #173

### DIFF
--- a/exercises/sum-of-multiples/sum-of-multiples.spec.js
+++ b/exercises/sum-of-multiples/sum-of-multiples.spec.js
@@ -2,23 +2,23 @@ var SumOfMultiples = require('./sum-of-multiples');
 
 describe('SumOfMultiples', function () {
   it('to 1', function () {
-    expect(SumOfMultiples().to(1)).toBe(0);
+    expect(SumOfMultiples([3, 5]).to(1)).toBe(0);
   });
 
   xit('to 3', function () {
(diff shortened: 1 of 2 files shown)


exercises/sum-of-multiples/sum-of-multiples.spec.js
```diff
@@ -2,23 +2,23 @@ var SumOfMultiples = require('./sum-of-multiples');
 
 describe('SumOfMultiples', function () {
   it('to 1', function () {
-    expect(SumOfMultiples().to(1)).toBe(0);
+    expect(SumOfMultiples([3,5]).to(1)).toBe(0);
```
petertseng: I know it's super minor, but it's possible this could be `[3, 5]` with an extra space if want to be consistent with the styles of all below lists, which do have spaces after their commas.

petertseng:
Yup, this is the right thing to do.

There are some whitespace changes that don't necessarily need to be in the same commit as the actual `[3, 5]` change, but I leave that to you to decide.

exercises/sum-of-multiples/sum-of-multiples.spec.js
matthewmorgan: Yes, I agree-- good catch.  I'll try to get to this tonight or first thing tomorrow.  Thanks!

PR #175, opened 2016-03-27
CustomSet: Adds test for eql method specifically

rrgayhart:
When going through the `custom-set` exercise, I felt like the first test covered both the .eql and the .delete methods and could cause some confusion.

This PR adds a test, matching the other test styles, to specifically target the `.eql` method prior to diving into testing `.delete`

### DIFF
--- a/exercises/custom-set/custom-set.spec.js
+++ b/exercises/custom-set/custom-set.spec.js
@@ -2,6 +2,18 @@ var CustomSet = require('./custom-set');
 
 describe('CustomSet', function() {
 
+  it('can compare equality of sets', function(){
+    var subject = new CustomSet([3, 2, 1]);
+    var duplicateSubject = new CustomSet([3, 2, 1]);
+    expect(subject.eql(duplicateSubject)).toBe(true);
+
+    var unOrderedDuplicateSubject = new CustomSet([3, 1, 2]);
+    expect(subject.eql(unOrderedDuplicateSubject)).toBe(true);
+
+    var nonDuplicateSubject = new CustomSet([2, 1]);
+    expect(subject.eql(nonDuplicateSubject)).toBe(false);
+  });
+
   it('can delete elements', function(){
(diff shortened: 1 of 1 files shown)


matthewmorgan:
@rrgayhart thanks for doing this.  Your PR made me look at the [test metadata](https://github.com/exercism/x-common/blob/master/custom-set.json) for this exercise on the x-common repo, and it looks like this track does not adhere very carefully (at all?) to that spec.

In the interest of getting this track aligned with the metadata, would you be willing to modify your PR to implement at least some of the test cases in the first series (lines 33-85 in the [metadata](https://github.com/exercism/x-common/blob/master/custom-set.json)?

If that is too much trouble then I understand of course.

rchavarria:
@rrgayhart, do you have any news on this? Your PR looks nice, but I agree with @matthewmorgan, if you could add some tests from the metadata (lines 33-85) it would improve this exercise.

If you need some help, @matthewmorgan and me will be glad to help you.

rrgayhart:
Thanks for reaching out and responding! I don't think I'll be able to respond/update this in a timely manner so I'll close the PR for now!

rchavarria:
I'm sorry to hear that. Your test really provide value to the exercise. You may didn't know about tests in x-common (for all languages track) so you couldn't focus on better tests.

If you find time, you could write some new tests, extracted from the exercise [metadata](https://github.com/exercism/x-common/blob/master/custom-set.json), for example:

From the exercise metadata, lines 42-47:

    {
        "description": "duplicates don't matter",
        "set1": [1, 1],
        "set2": [1],
        "expected": true
     }

That spec would translate into the following test in the file you modifyed on this PR:

    it('duplicates do not matter', function () {
        var set1 = new CustomSet([1, 1]);
        var set2 = new CustomSet([1]);

        expect(set1.eql(set2)).toBe(true);
    });

That's all. I'd love receiving a PR from you. Thanks for your work.

PR #178, opened 2016-04-04
fix `sorted is not defined` error.

GantMan:
for those of us with a bit more strict Javascript compilation.

image

### DIFF
--- a/exercises/grade-school/grade-school.spec.js
+++ b/exercises/grade-school/grade-school.spec.js
@@ -28,7 +28,7 @@ describe('School', function() {
   xit('adding students to different grades adds them to the roster', function() {
     school.add('Chelsea',3);
     school.add('Logan',7);
-    var expectedDb = { 3 : [ 'Chelsea' ], 7 : [ 'Logan'] };
+    var expectedDb = { 3 : [ 'Chelsea' ], 7 : [ 'Logan' ] };
     expect(school.roster()).toEqual(expectedDb);
   });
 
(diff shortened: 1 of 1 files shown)


matthewmorgan:
@GantMan thanks for doing this.  Good argument for getting linting set up for this track...

One minor request: stylistically, all the other array definitions have spaces between the brackets and the values.  Would you mind changing that so we stay consistent?  If you don't have time let me know and I can do it.  Thanks again.

GantMan:
I agree wholeheartedly on the linting.  That would catch most issues nicely.
**RE: Spaces** - I believe I know what you're saying.   I can do that right now.

GantMan:
I believe this is what you wanted.   Please review.

rchavarria:
I think @matthewmorgan refers more to

```
var sorted = { 3 : [ 'Kyle' ], 4 : [ 'Christopher', 'Jennifer' ], 6 : [ 'Kareem' ] };
```

Notice `3 : ['K` and `3 : [ 'K`.

By the way. There is a similar *typo* with whitespace in the line 31. I know it's not your fault, but, because it's a similar change, would you mind to modify it?

Thanks a lot!

GantMan:
Ahh, now I see @rchavarria - Fixed! (line 31 as well).

matthewmorgan:
@GantMan thanks!  one last thing-- when possible, we like to have one commit per PR.  Can you squash the 3 commits to one?  If you don't have time let me know and I can do it.  No problem either way.

GantMan:
sorry for the noise, had to remember how to squash commits.   PR should be good to go!

matthewmorgan:
@GantMan @rchavarria thanks!  Better and better...

Issue #109, opened 2015-09-15
Wordcount is different from python or java


jmluy:
Tests for word count is different in that it doesn't normalize to lowercase or even strip special characters.  It's not mentioned in the readme file but It might be better to have similar tests.

kytrinyx:
Yeah, the word count exercise has evolved a bit over time. This might be a good time to extract the best test cases into a [shared set of metadata](https://github.com/exercism/todo/issues/7) so that the various tracks can [write generators](https://github.com/exercism/todo/issues/8)

wobh:
It seems to me that, ideally, the word count test should be indifferent to the method of normalizing. Just did a PR to the xpython about this.

kytrinyx:
Agreed--I really like that change.

PR #179, opened 2016-04-04
Update tests and example to be case-agnostic

matthewmorgan:
Addresses #109

### DIFF
--- a/exercises/word-count/word-count.spec.js
+++ b/exercises/word-count/word-count.spec.js
@@ -28,7 +28,7 @@ describe('count()', function() {
     expect(words.count('testing 1 2 testing')).toEqual(expectedCounts);
   });
 
-  xit('respects case', function() {
+  xit('normalizes to lowercase', function() {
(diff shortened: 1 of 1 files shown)


exercises/word-count/word-count.spec.js
```diff
@@ -29,12 +29,12 @@ describe('count()', function() {
   });
 
   xit('respects case', function() {
```
rchavarria: I think this description does not reflect what the spec does now. What about *normalizes to lower case*? @matthewmorgan, if you agree, I can do the change while merging the PR.

matthewmorgan: Yes, you are right. Do you mind changing it? Thanks!

rchavarria:
It appears as *Closed* instead of *Merged* because I introduced a [new commit](https://github.com/exercism/xjavascript/commit/30b4a7e3bcfc372dac6af24490d6437dd7e1bf3d).

PR #183, opened 2016-04-07
Replace "multibillionaire" with "oxyphenbutazone"

matthewmorgan:
As described in #180


rchavarria:
Thanks mate! You could have merged the PR, it's very simple. But it feels good to merge it, so no problem.

PR #188, opened 2016-04-20
Clock: synchronize tests with canonical ones

rchavarria:
Addresses #184 

Quite similar to PR in [xecmascript#218](https://github.com/exercism/xecmascript/pull/218).

### DIFF
--- a/exercises/clock/example.js
+++ b/exercises/clock/example.js
@@ -1,11 +1,26 @@
 exports.at = at;
 
+var HOURS_IN_A_DAY = 24;
+var MINUTES_IN_AN_HOUR = 60;
+var MINUTES_IN_A_DAY = HOURS_IN_A_DAY * MINUTES_IN_AN_HOUR;
+var MILLIS_IN_A_MINUTE = 60 * 1000;
+var MILLIS_IN_AN_HOUR = MINUTES_IN_AN_HOUR * MILLIS_IN_A_MINUTE;
+var MILLIS_IN_A_DAY = HOURS_IN_A_DAY * MILLIS_IN_AN_HOUR;
+
+function makePositive(time, maxValue) {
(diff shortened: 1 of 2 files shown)


exercises/clock/example.js
```diff
@@ -1,11 +1,25 @@
...
+
 function at(hours, minutes) {
-  var min = 1000 * 60;
-  var hr = min * 60;
+  hours = makePositive(hours, HOURS_IN_A_DAY);
+  minutes = makePositive(minutes, MINUTES_IN_A_DAY);
 
   var clock = {};
-  var value = (~~hours * hr) + (~~minutes * min);
+  var value = (~~hours * MILLIS_IN_AN_HOUR) + (~~minutes * MILLIS_IN_A_MINUTE);
```
matthewmorgan: I'm wondering what the `~~` is doing here.  I've never used it, but I see it's equivalent to `Math.floor()`.  Do we need `Math.floor()` here?

rchavarria: I thought `~~` was a weird way to convert a value into an integer (well, Number in JS), but it's probably equivalent to `Math.floor()`.

I'll try to remove it, and I'll come back. Wait a minute :smiley:

rchavarria: Nice catch @matthewmorgan!!

They were not neccessary, but I needed to include

    minutes = minutes || 0;

at line 17 because the `minutes` parameter was optional, so `~~` helped to convert it to a number. 

Now, with your suggestion is much cleaner, thanks!!

matthewmorgan:
Thanks @rchavarria !  You are the [energizer bunny](http://i.giphy.com/l4hLxqjeiKovUYuIw.gif) of xjavascript!

rchavarria:
LOL

Very fun!

PR #189, opened 2016-04-22
What do you think about this hack?

yvan-sraka:
We use [GitHub](https://github.com/) and [Travis CI](https://travis-ci.org/) to let students of [Simplon.co](http://simplon.co/) practice both JavaScript and GitHub Test Driven Development Workflow.

To check a solution, just open a `pull-request` on the main repository, and let Travis CI do his job :wink: !


rchavarria:
Thanks for the PR and your interest in improving the code so that others can learn from it.

Actually, this repo has configured Travis CI to run tests on each commit or PR. But as a whole repo, not for each individual exercise. This repo is part of the [exercism.io](http://exercism.io/) platform, and each individual exercise is supposed to be solved in that platform (it's not just solving an exercise, it's the social interaction you get with that: technical feedback, comments from other developers,...).

So, for me, adding a Travis config file for each exercise is not aligned with the goal of this repo, sorry. Maybe, other @exercism/track-maintainers have a word on this. I can't take a decission myself.

Additionally, adding FizzBuzz exercise is a nice idea. It's a well known exercise to learn TDD and it causes some interesting discussions. Sorry, I'm sure there is a document describing how to add a new exercise to a language track (like this repo), but I can't find it (@matthewmorgan can you help me?)

petertseng:
Hmm, Travis would [look for a .travis.yml in the root of the repository](https://docs.travis-ci.com/user/customizing-the-build/), right?

So does that mean: For this to work, every exercise needs to be its own separate repository? It would be possible to [do that with git-subtree](http://stackoverflow.com/questions/920165/howto-extract-a-git-subdirectory-and-make-a-submodule-out-of-it) if that is desired. I wonder if it would be difficult to manage all the separate repos that would result! It seems it can be less work to use only one .travis.yml for all of xjavascript track, since the way to test the each problem is similar.

> I'm sure there is a document describing how to add a new exercise to a language track

https://github.com/exercism/x-common/blob/master/CONTRIBUTING.md#implementing-a-completely-new-exercise

NobbZ:
I think, that testing the repository as a whole, does improve the quality
of commits.

These tests do involve checking for proper configuration of the track as
well as some level of consistency. Especially when upgrading meta stuff
like updating the test runners version needs the complete run.

Also splitting this up would make a track unmaintanable as Peter said.

I do see only pros in testing everything.

Peter Tseng <notifications@github.com> schrieb am Mi., 27. Apr. 2016, 10:20:

yvan-sraka:
Thank you for all your comments, I agree that my `pull-request` is a bit _messy_ :scream:..
So I can create a special `branch` for adding the `FizzBuzz` exercise, and keep the other commits of the `fork` for my personal use :wink:!

**P.S.** You can find the repository we create for this hack [here](https://github.com/simplonco/exercism.io).

kotp:
The fizzbuzz exercise is effectively the Raindrops exercise, if curious.

ryanplusplus:
Take a look at https://github.com/exercism/xlua. We're using Travis and a simple shell script to test each exercise independently. While not the most sophisticated setup in the world, it lets us make sure everything works without making the track unmaintainable.

yvan-sraka:
```shell
$ jasmine-node *.spec.js --verbose --captureExceptions
count() - 8 ms
    counts one word - 6 ms

Failures:
  1) count() counts one word
   Message:
     Expected 1 to equal { word : 1 }.
   Stacktrace:
     Error: Expected 1 to equal { word : 1 }.
    at null.<anonymous> (/home/travis/build/simplonco/word-count/word-count.spec.js:8:33)

Finished in 0.013 seconds
1 test, 1 assertion, 1 failure, 0 skipped
```
Also take a look at https://github.com/jasmine/jasmine, to run tests locally :wink: !

matthewmorgan:
@yvan-sraka @kotp @rchavarria This exercise is pretty much the 'raindrops' exercise, I agree.  For that reason, I don't think it makes sense to add it.

But, @yvan-sraka if you did want to add an exercise in the future (which would be awesome) there is a full list of exercism exercises (@kytrinyx where is that list?  I thought it was in x-common but I can't find it), and each language track repo has a list of what is implemented in that language.  

So, you would either:
1) Add a new exercise definition to the master list (one PR), then add it to the track (a second PR), or
2) Find an exercise on the master list that's not on the track, and add it with a PR.

kotp:
No need to really stick to the available list either... but it is good to know what is here so you don't duplicate it.  The way to find the missing exercises in language tracks is covered in https://github.com/exercism/discussions/issues/1

kytrinyx:
To echo what the others have said: Thank you so much for jumping in with suggestions and code! It's much appreciated.

I agree that fizzbuzz would be too similar to raindrops, in this case, to make sense to add it.

I also don't want to add travis config files to each individual exercise, but if you see a need for a way to test things that we don't currently have, definitely open an issue, and we can figure out what the pain point is, and maybe we'll find a way to fix that which is compatible with what we're doing.

The full list of existing problems is in github.com/exercism/x-common (all the *.md files are problem descriptions). We're always looking to add more exercises, so suggestions, pull requests, ideas, etc are all welcome. Sometimes we'll find that an exercise is too similar to one that we already have, but the discussions often spark interesting new things anyway, so it's never wasted.

I'm going to go ahead and close this. Thanks again for taking the time to submit it!

PR #191, opened 2016-04-29
Make the test instructions more beginner-friendly

kytrinyx:
The github-flavored-markdown section of SETUP using the fenceposts
is completely non-obvious if you aren't familiar with markdown.

I just watched a newbie try to type in the backticks and "bash"
along with the dollar-sign.

This removes some of the clutter.

### DIFF
--- a/SETUP.md
+++ b/SETUP.md
@@ -9,9 +9,7 @@ http://exercism.io/languages/javascript
 
 Execute the tests with:
 
-```bash
-$ jasmine-node .
-```
+    jasmine-node .
 
 In many test suites all but the first test have been skipped.
 


rchavarria:
I've found more occurrences of those fenceposts on:

- `docs/INSTALLATION.md`
- `docs/TESTS.md`
- `SETUP.md`

Don't worry, I'll take care of them. I'll replace `$` occurrences as well.

kytrinyx:
The ones in docs aren't a problem, because they're not served as part of the README. People will look at them on the site, not on their machine using their text editor.

I'm not sure about the dollar sign... Maybe we need to have a place where we explain conventions.

rchavarria:
I've seen the dollar sign in some tutorials, they mimic the Unix prompt. But, if you want to copy and paste a couple of commands, it gets in the way. 

There is no problem, and it won't be a hard job to find and replace occurrences. No problem.

kytrinyx:
Yeah, let's just get rid of it.

PR #192, opened 2016-04-29
Clarify hello-world stub file comments

kytrinyx:
I removed 'only' because it doesn't seem to serve a purpose.
I switched SKELETON to stub because we don't need to shout, and we use the term stub file elsewhere
on the site.
Lastly, I changed the file name to match the actual file name in the exercise.


rchavarria:
Yep, no need to shout.
Nice catch the outdated file name.
Thanks!!

PR #193, opened 2016-04-29
Add a basic tutorial for hello-world

kytrinyx:
If you've only ever written javascript for the browser, and you've
never done any TDD, then this exercise has a lot of barriers to success.

This tutorial might help get people started.

I'm not sure exactly what needs to go into the tutorial. This is based on watching a code newbie try to figure out how to get started. They've done some tutorials where they write code in the browser, but never any TDD or node.

### DIFF
--- a/exercises/hello-world/HINTS.md
+++ b/exercises/hello-world/HINTS.md
@@ -1,0 +1,102 @@
+## Tutorial
+
+This exercise has two files:
+
+- hello-world.js
+- hello-world.spec.js
+
+The first file is where you will write your code.
+The second is where the tests are defined.
+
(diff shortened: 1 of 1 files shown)


rchavarria:
Very, very nice. And very, very detailed. Just perfect for beginners.

I think the `HINTS.md` file is merged in the exercise `README` file so that the user is going to read it, right?

kytrinyx:
Yeah, HINTS.md and SETUP.md both get included in the README. SETUP is for the track-level stuff and gets included in all the READMEs, whereas HINTS only gets included for the specific exercise.

PR #195, opened 2016-04-30
Remove clutter from documentation

rchavarria:
Remove clutter from markdown text to make it simpler for beginners.

This PR complements #191 because there were more occurrences of fenceposts.


matthewmorgan:
@rchavarria Thanks!

PR #198, opened 2016-05-04
Isogram: tests and example

ecrmnn:
Tests and example for Isogram exercise

### DIFF
--- a/config.json
+++ b/config.json
@@ -6,14 +6,15 @@
   "test_pattern": ".*[.]spec[.]js$",
   "problems": [
     "hello-world",
-		"leap",
+    "leap",
     "hamming",
-		"rna-transcription",
-		"bob",
-		"gigasecond",
-		"word-count",
+    "rna-transcription",
+    "bob",
+    "gigasecond",
+    "word-count",
+    "isogram",
     "pangram",
-		"beer-song",
+    "beer-song",
     "phone-number",
     "anagram",
     "food-chain",
@@ -69,7 +70,7 @@
     "two-bucket"
   ],
   "deprecated": [
-		"nucleotide-count",
+    "nucleotide-count",
     "point-mutations"
   ],
   "ignored": [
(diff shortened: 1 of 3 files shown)


matthewmorgan:
@ecrmnn thanks for doing this!

I have a couple points of feedback:
1) it would be good to have your example (and probably a test or two) that test for casing on the isograms.  EG, shouldn't 'Kicker' fail the test?  Right now I think it would pass BC the two k's are different cases.
2) The `config.json` file controls the order in which the exercises are served to users.  This exercise is excellent for JavaScript beginners, but is not as complex to solve as the other exercises at the end of the track, it should be placed closer to the beginning, near something like `word-count` or `pangram`.

Would you be able to make those changes?  Let me know if you need help, I'm happy to to assist.

ecrmnn:
@matthewmorgan Thanks for your feedback.

1. Changed test from "alphabet" to "Alphabet". This would have passed before, now it's fixed.
2. I didn't realize it was the order the exercises were served in. Sorry about that. Moved *isogram* between ``word-count`` and ``pangram``.

matthewmorgan:
Great!  Thanks.  One last thing: the custom here is to squash all the commits for a given PR, so there's only one commit in the master history once it's merged.

Are you comfortable squashing this or would you like help?

ecrmnn:
@matthewmorgan All good now?

matthewmorgan:
oops-- one more thing: all tests but the first should be changed to `xit` so they don't run out of the box.  Otherwise it looks great.

ecrmnn:
Oups, fixed

matthewmorgan:
Awesome.  Thanks @ecrmnn !

ecrmnn:
No problem 😄 [Exercise files](https://github.com/exercism/x-common/pull/247) should probably be merged too in order for this to work

matthewmorgan:
Thanks for the reminder!  I forgot that detail 😊

PR #199, opened 2016-05-05
word-count: use 'constructor' as a reserved word

ireddick:
As 'prototype' isn't a standard property on an object's prototype,
the test didn't actually check the case of dealing with a collision

The 'constructor' property is a standard

see: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object
and: http://www.ecma-international.org/ecma-262/5.1/#sec-15.2.4

It's possible that this will break existing solutions - the reference solution is immune though

### DIFF
--- a/exercises/word-count/word-count.spec.js
+++ b/exercises/word-count/word-count.spec.js
@@ -59,7 +59,7 @@ describe('count()', function() {
   });
 
   xit('handles properties that exist on Object’s prototype', function() {
-    var expectedCounts = { reserved: 1, words : 1, like :1,  prototype: 1, and : 1, tostring: 1,  'ok?': 1};
-    expect(words.count('reserved words like prototype and toString ok?')).toEqual(expectedCounts);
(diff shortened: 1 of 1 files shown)


rchavarria:
You're right. Even the test says

> ... that exist on Object’s prototype

So `prototype` itself doesn't seem to be the right property, while `constructor` and `toString` do.

Thanks for you contribution, well done!!

rchavarria:
Don't worry for existing solutions. They may break, but some exercises add new tests, or remove old ones, and that also makes them break. That can generate new conversations on exercism.io, and that's the goal of the platform: [It's about the conversations!](http://tinyletter.com/exercism/letters/exercism-it-s-about-the-conversations).

ireddick:
This will also need to be changed on the ecmascript track - shall I raise an issue?

rchavarria:
Yes please!

ireddick:
raised: https://github.com/exercism/xecmascript/issues/224

remembered to quote the ES6 spec this time!

PR #274, opened 2016-05-13
linked-list: add missing expectation

### LINKED ISSUE #270: linked-list: missing expectation
The test 'deletes the only element' is missing an expectation.

Needs an expectation added, or the test deleted. Maybe check with other language tracks to see if the test exists there, and what is being asserted on?

ireddick:
Fixes #270

This is the same expectation as that on the ecmascript track


rchavarria:
Thanks a lot for this PR.

PR #285, opened 2016-06-10
Update custom set test order

### LINKED ISSUE #283: Updated tests for the Custom Set problem
In order to reduce the amount of code required to pass incremental tests (assuming that users pass tests starting from the top), the order of the tests was modified slightly.

Since this track implements Custom Set, please take a look at the new custom-set.json file and see if your track should update its tests.

If you do need to update your tests, please refer to this issue in your PR. That helps us see which tracks still need to update their tests.

If your track is already up to date, go ahead and close this issue.

More details on this change are available in https://github.com/exercism/x-common/pull/257.

ireddick:
fixes #283

Note that I've done this as 2 commits:

Commit 1 does not break existing solutions and does as much as possible to conform to the canonical tests. This also uncovered a bug in the example solution, which this fixes.

Commit 2 will break existing solutions, as it changes #empty to have the meaning from the canonical tests, and renames the existing method to #clear

The non-canonical tests for this track have been left (they are positioned after the canonical tests)

I've stayed with the style of the test suite to keep things consistent

Happy to discuss, squash or change any of this!

### DIFF
--- a/exercises/custom-set/custom-set.spec.js
+++ b/exercises/custom-set/custom-set.spec.js
@@ -2,6 +2,151 @@ var CustomSet = require('./custom-set');
 
 describe('CustomSet', function() {
 
+  xit('can test for no members', function() {
+    var actual = new CustomSet().empty();
+    expect(actual).toBe(true);
+
+    var actual2 = new CustomSet([1]).empty();
+    expect(actual2).toBe(false);
+  });
+
+  xit('can test for a member', function() {
+    var actual = new CustomSet().contains(1);
+    expect(actual).toBe(false);
+
+    var actual2 = new CustomSet([1, 2, 3]).contains(1);
+    expect(actual2).toBe(true);
+
+    var actual3 = new CustomSet([1, 2, 3]).contains(4);
+    expect(actual3).toBe(false);
+  });
+
+  xit('can test for subsets', function() {
+    var actual = new CustomSet().subset(new CustomSet());
+    expect(actual).toBe(true);
+
+    var actual2 = new CustomSet([1]).subset(new CustomSet());
+    expect(actual2).toBe(true);
+
+    var actual3 = new CustomSet().subset(new CustomSet([1]));
+    expect(actual3).toBe(false);
+
+    var actual4 = new CustomSet([1, 2, 3]).subset(new CustomSet([1, 2, 3]));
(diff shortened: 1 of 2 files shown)


ireddick:
this isn't quite right - see comment on issue #283

ireddick:
see issue #283 for discussion

rchavarria:
Nice! Great job! I think changes are ok to be merged. I'll wait a little bit for you to squash your commits and it'll be merged.

rchavarria:
I would like to split each test case, to reflect better the test cases exposed in the canonical tests *document* [`custom-set.json`](https://github.com/exercism/x-common/blob/master/custom-set.json)

Let me explain: in the canonical tests, there are three *cases* for the `contains` method, with these descriptions:

- nothing is contained in an empty set
- when the element is in the set
- when the element is not in the set

But, in the test suite, there is only a spec (`it` block) that covers the three cases:

```
xit('can test for a member', function() {
    var actual = new CustomSet().contains(1);
    expect(actual).toBe(false);

    var actual2 = new CustomSet([1, 2, 3]).contains(1);
    expect(actual2).toBe(true);

    var actual3 = new CustomSet([1, 2, 3]).contains(4);
    expect(actual3).toBe(false);
});
```

I think the test suite will be clearer for exercism users if we split `it` blocks in this way:

```
xit('nothing is contained in an empty set', function() {
    var actual = new CustomSet().contains(1);
    expect(actual).toBe(false);
});
xit('when the element is in the set', function() {
    var actual = new CustomSet([1, 2, 3]).contains(1);
    expect(actual).toBe(true);
});
xit('when the element is not in the set', function() {
    var actual = new CustomSet([1, 2, 3]).contains(4);
    expect(actual).toBe(false);
});
```

But, test cases weren't splitted in the begginning, so I think this code is good to be merged. You've made a great job. Splitting the test cases could be a totally different PR. It'd be nice if you can work on it, if you want to.

ireddick:
Yes, I came to the same conclusion about the need to split the test cases - I also think it's out of scope of the original issue. I'll raise an issue so that it's visible, and I'm more than happy to work on it.

ireddick:
squashed and ready to go

rchavarria:
Great!!!!

Thanks so much @ireddick

PR #291, opened 2016-07-01
Add track icon

wolverineks:


rchavarria:
Thanks!! Lots of them!

Would you like to do same in the xecmascript track ([xecmascript-231](https://github.com/exercism/xecmascript/issues/231))? It's not mandatory, it's just if you want.

PR #293, opened 2016-07-07
Add new test in the Isogram exercise

rchavarria:
éléphant is not an isogram, but some solutions using a
simple regexp replace of non-word characters will fail
at it.

Addresses #287


rchavarria:
@matthewmorgan, can you take a quick look at this? Just say ok or not, and I'll take care of the rest.

rchavarria:
Well, the PR was so small and inocent that I merge it myself.

PR #295, opened 2016-07-22
custom-set: correct initially enabled test

ireddick:
The first test should be enabled, all others should be disabled

This was missed in the re-ordering of the tests that I did a few weeks ago.


rchavarria:
Ooops! I didn't realize that. Nice catch!

This PR is ready to be merged, but I need to find time to do it. 
Thanks for the contribution @ireddick

rchavarria:
Done!

Thanks @ireddick !!

PR #296, opened 2016-07-23
modified gigasecond assignment to use a getDate method

JamesBedont:
Given a user writes a solution to the Gigasecond problem using the Pseudoclassical method:
```
var Gigasecond = function(date) {
  this.dob = date;
};

Gigasecond.prototype.date = function () {
  var gigasecondDate = new Date(this.dob.getTime() + 1000000000000);
  return gigasecondDate;
};

module.exports = Gigasecond;
```

the spec will fail on line 8 with the error of **TypeError: gs.date is not a function**.

line 8 of gigasecond.spec.js
```
expect(gs.date()).toEqual(expectedDate);
```

I'm not 100% sure why but if all occurrences of *gs.date()* are replaced with something like *gs.getDate()* this problem doesn't occur. I assume it is because *date* is a reserved word in JavaScript or something similar. I would suggest making this change because users are given stub files from the first exercise using the Pseudoclassical method of writing JavaScript; so when it fails at this exercise I can see people getting frustrated.

[this would be a clear example of user getting burned on this.](http://exercism.io/submissions/3dbb2f7491944d048ab4eff7cddb24f6) You can see that he tried the way that has been shown for all previous methods commented it out then did it another way.

### DIFF
--- a/exercises/gigasecond/example.js
+++ b/exercises/gigasecond/example.js
@@ -3,7 +3,7 @@ function Gigasecond(dateOfBirth) {
 
   this.dateOfBirth = dateOfBirth;
 
-  this.date = function() {
+  this.getDate = function() {
     var gigasecondDate = new Date(this.dateOfBirth.getTime() + 1000000000000);
     return gigasecondDate;
   };
--- a/exercises/gigasecond/gigasecond.spec.js
+++ b/exercises/gigasecond/gigasecond.spec.js
@@ -5,27 +5,25 @@ describe('Gigasecond', function() {
   it('tells a gigasecond anniversary since midnight', function() {
     var gs = new Gigasecond(new Date(Date.UTC(2015, 8, 14)));
     var expectedDate = new Date(Date.UTC(2047, 4, 23, 1, 46, 40));
-    expect(gs.date()).toEqual(expectedDate);
+    expect(gs.getDate()).toEqual(expectedDate);
   });
 
   xit('tells the anniversary is next day when you are born at night', function() {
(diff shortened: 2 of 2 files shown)


rchavarria:
I don't know why it fails. [My own solution on exercism](http://exercism.io/submissions/54b7f4d8380347efa96836acfba8f6e1) is quite similar to the example you show, and it worked when I submitted it (there might have been changes since then, so my solution may fail today):

```
function Gigasecond (birthday) {
  this.birthday = birthday;
}

Gigasecond.prototype.date = function () {
  // ...
}

module.exports = Gigasecond;
```

And I think the problem is not that `date` is a reserved word in JavaScript (I think it's not a reserved word, but I don't know for sure, I should check) because all specs are running and passing on a continuous integration server: https://travis-ci.org/exercism/xjavascript/builds/

Before closing or merging this PR, give me some time to test the code you provide with the current specs. I'll post back here my findings.

Thanks for your work. I see you want to help users, so be patient and give me some time to check this.

JamesBedont:
So I have gone back and re-run the tests and now they are all passing with no issue around the date method. unlike when I first made this PR.

When I made this PR I had probably run the tests at least 20 times and it always failed with a type error. I'm not sure what is different about my environment today as to my knowledge nothing has changed.

You can close this PR as far as I'm concerned. Not sure what went wrong for me.

rchavarria:
Ok, I'll close this.

Anyway, many thanks for taking your time to work on this, to explain it, to re-check it and so on. Thanks a lot!!

Issue #289, opened 2016-06-25
Improve documentation about how to run tests


rchavarria:
As discussed in #284, some users find it difficult to start using this track and start running tests.

Improve documentation on `/docs/TEST.md` by describing more or less this idea:

> In the command line, go to the directory where exercism place test files (`<EXERCISM_HOME_DIR>/<TRACK_ID>/<EXERCISE>`), and run the `jasmine-node .` command.

It would be nice if an example is provided.

rchavarria:
After #297 and #330, I think this can be closed.

If there is no problem, I'll close it in a few days.

Issue #284, opened 2016-06-03
installing json and jasmine-node package


jsoranno:
I've needed to Google a lot to figure out why I couldn't run the test for the Hello World tutorial. I'm running on OSX 10.11.4 and I needed to do the following:
Installed Homebrew
$ brew update && brew install exercism
$ exercism --version
$ exercism configure --key=_(mine)_
$ exercism fetch javascript hello-world
$ jasmine-node .     // didn't work
$ sudo npm install jasmine-node -g
$ jasmine-node . // tried again and it spit back
(node) sys is deprecated. Use util instead. // I have no idea what that means, so I kept Googling
$ npm init // Then I created some json package that I don't completely understand... after working through the setup wizard
$ npm install jasmine-node --save
$ cd exercism
$ jasmine-node . // FINALLY!!! 

Is there a way to add in the $cd exercism instructions after all of the install instructions, but before doing the first test? Should I create a pull request? Or maybe this exists somewhere and I didn't come across it...

rchavarria:
I understand, very frustrating.

The recommended way of running tests is explicitly name the spec file (file with the tests):

    $ jasmine-node hello-world.spec.js

I don't know if just `jasmine-node .` does the work. Apparently, for your `// FINALLY!!` comment it works.

I'm not sure, but following the steps you describe, I think you're missing a very important one. Navigate to the folder where the exercise is fetched. When you run

    $ exercism fetch javascript hello-world

It fetches some files and put them under

    <EXERCISM_HOME_FOLDER>/javascript/hello-world

where `<EXERCISM_HOME_FOLDER>` is `~/exercism` by default (see [command-line interface documentation](http://exercism.io/cli) to learn more about the `exercism` tool, how to configure it, ...).

So, after fetching the exercise, you should navigate to `~/exercism/javascript/hello-world`. Then, if you installed `jasmine-node` globally (with the `-g` option of `npm`) you should be able to run tests with `jasmine-node hello-world.spec.js`.

@kytrinyx I'm not able to find a reference in the documentation about navigating to the folder where `exercism` tool put fetched files. Do you think it would be necessary to state it somewhere? I'd love to help.

kytrinyx:
It depends on the language you're working in. In Java there's a whole directory tree that gets fetched. Maybe we can suggest that people navigate to the folder of the exercise, so `<EXERCISM_HOME_DIR>/<TRACK_ID>/<EXERCISE>` and work from there. I *think* that should work for all the languages.

[...]

PR #297, opened 2016-07-28
Elaborate a bit more on how to run the tests for the first time

rchavarria:
This PR adds a more detailed explanation about how a user should run the tests for the first time, because some users didn't find it easy to start running them (see #284).

Addresses #289

### DIFF
--- a/docs/TESTS.md
+++ b/docs/TESTS.md
@@ -1,6 +1,9 @@
-Execute the tests with:
+After configuring `exercism` command-line client and fetching your first JavaScript exercise (see [command-line interface overview](http://exercism.io/cli)) is time to run some tests. Move to the folder where that exercise's files are located (a path similar to `<EXERCISM_HOME_DIR>/<TRACK_ID>/<EXERCISE>`) and run the tests with the `jasmine-node` command you should have installed on *Installing JavaScript* step:
 
+    cd ~/exercism/javascript/bob
     jasmine-node bob_test.spec.js
+
+*Note that `~/exercism` is the default folder for `EXERCISM_HOME_DIR`. Be sure you use your configured folder for it*
 
 ## Making Your First Node Module
 


rchavarria:
@kytrinyx Let me know if you think this PR addresses the issue described on #284 correctly. If so, I would take care and merge it.

docs/TESTS.md
```diff
@@ -1,7 +1,18 @@
-Execute the tests with:
+After configuring `exercism` command-line client with:
 
+    exercism configure --key=YOUR_EXERCISM_KEY
+
+and fetching your first JavaScript exercise with:
+
+    exercism fetch javascript bob
```
kytrinyx: I worry about duplicating instructions about configuration and fetching (though we probably already do), since if we change it it's going to be pretty hard to remember to get all the other spots.

I do like how it refers to both configuring and fetching, though. That helps orient the person in space and time.

docs/TESTS.md
```diff
@@ -1,7 +1,18 @@
...
+
+is time to run some tests. Move to the folder where that exercise's files are located (a path similar to `<EXERCISM_HOME_DIR>/<TRACK_ID>/<EXERCISE>`) and run the tests with the `jasmine-node` command you should have installed on [Installing JavaScript](http://exercism.io/languages/javascript#installing) step:
+
```
kytrinyx: Yeah, I think this is good.

kytrinyx: That said... I don't think the '#installing' thing will work on exercism :(

rchavarria:
1. Instead of duplicating configuration and fetching instructions I just linked to the [CLI overview](http://exercism.io/cli). BTW, I miss how to fetch a problem on that overview. I can create a issue and write a couple of paragraphs to fix it if you want.
2. I removed the link ending with `#installing`. I tried and it doesn't work. Users should read the text in this PR having installing JavaScript instructions on the left side of the screen, so a link shouldn't be needed.

kytrinyx:
> I can create a issue and write a couple of paragraphs to fix it if you want.

Yes please!

kytrinyx:
I think this is good to go.

matthewmorgan:
@rchavarria @kytrinyx Thanks!

PR #298, opened 2016-08-08
Adding two more test cases for binary-search

junedev:
When doing this exercise I had some troubles getting the indexes correct. I was passing the tests that were provided but my program did not work correctly for entries at the very beginning or end of the array e.g. because of some off-by-one errors.
That's why I am proposing to add two test cases here so people don't have to manually add them like I did.

### DIFF
--- a/exercises/binary-search/binary-search.spec.js
+++ b/exercises/binary-search/binary-search.spec.js
@@ -14,8 +14,16 @@ describe('BinarySearch', function() {
     expect(Array.isArray(validBinarySearch.array)).toEqual(true);
   });
 
-  xit('should find the correct index of an included value', function() {
+  xit('should find the correct index of an included value in the middle of the array', function() {
     expect(new BinarySearch(sortedArray).indexOf(3)).toEqual(2);
+  });
+
+  xit('should find the correct index of an included value at the beginning of the array', function() {
+    expect(new BinarySearch(sortedArray).indexOf(1)).toEqual(0);
+  });
+
+  xit('should find the correct index of an included value at the end of the array', function() {
+    expect(new BinarySearch(sortedArray).indexOf(6)).toEqual(5);
   });
 
   xit('should search the middle of the array', function() {
@@ -26,4 +34,3 @@ describe('BinarySearch', function() {
     expect(new BinarySearch(sortedArray).indexOf(10)).toEqual(-1);
   });
 });
-


matthewmorgan:
@junedev LGTM!  Thanks!  @rchavarria ?

junedev:
Thanks for having a look. Once this is merged here I will also make a pull request for the EcmaScript track as the exercise seems to be exactly the same there.

matthewmorgan:
Yeah, the other thing that might make sense is to add metadata about what tests and results we want to see defined for all the tracks-- right now, it's not defined at all for this exercise, but if we were to have a file like that, It should include your tests.

These are in the x-common repo, for example: [binary](https://github.com/exercism/x-common/blob/master/binary.json)

junedev:
Thanks for the explanation. I will create the missing files for the common repository (probably next week though). Is there an easy way to see which other exercises might be missing in x-common or do I have to check one by one?

rchavarria:
@junedev I fear you have to check one by one :(

rchavarria:
About the PR, it's ok. It follows other tests' structure and all tests keep passing.

I'm gonna merge this.

matthewmorgan:
@rchavarria @junedev I think the question @junedev was asking is about if any of the metadata for binary-search is defined x-common.  Right now, there is no `config.json` file, so none of them are defined.

@kytrinyx has an automated script to create an issue for each track to updates its tests to conform once we have some agreement on what tests are wanted.

junedev:
@matthewmorgan: @rchavarria understood correctly. I was asking about other exercises besides binary-search ... just in case I have some more time. Then I could also add other missing common files not just the ones for binary-search.

matthewmorgan:
OK!  got it.  You are ambitious 😸

kytrinyx:
> Is there an easy way to see which other exercises might be missing in x-common or do I have to check one by one?

@junedev We have this endpoint which might be what you're looking for: http://x.exercism.io/v3/problems/data-todos

junedev:
@kytrinyx I was looking for a list of all problems that are missing the shared input/output json data. After a bit of strolling around in the exercism repos and issues I finally found it here: https://github.com/exercism/todo/issues
So I am good to go. :smile:

kytrinyx:
Ahh, nice! Glad you're all set.

PR #300, opened 2016-08-20
Update config.json to match new specification

Roshanjossey:
> This addresses issue #299
> https://github.com/exercism/xjavascript/issues/299

This PR addresses this task of the issue mentioned
>Add the exercises key with the array of objects, where difficulty is 1 and topics is empty.

### DIFF
--- a/config.json
+++ b/config.json
@@ -71,6 +71,398 @@
     "two-bucket",
     "bowling"
   ],
+  "exercises": [
+    {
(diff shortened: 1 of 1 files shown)


kytrinyx:
Thank you so much, @Roshanjossey!

For reference, this is phase 1 as described in #299 - the next phases will adjust the difficulty level and add topics.

PR #304, opened 2016-10-16
beer-song: Add test case for 2 bottles.

nemtlaur:
An isolated test of the 2 bottle case would help highlight that it is a special case. If the 2 bottle case is not called out when 1 and 0 are, the error will not be caught until the final test.


rchavarria:
I'm not sure it provides as much value as `0`, `1` or arbitrary number of bottles, but I agree with you. Handling `2` bottles is a bit different that handling `3` or `8`.

I can't currently merge your PR, but it'll be merged very soon. I won't forget it.

Thanks so much for your contribution (code and new ideas), it's very much appreciated.

PR #308, opened 2016-11-30
Added all-your-base exercise to JavaScript track

javaeeeee:
Added all-your-base exercise to the JavaScript track

### DIFF
--- a/exercises/all-your-base/all-your-base.spec.js
+++ b/exercises/all-your-base/all-your-base.spec.js
@@ -1,0 +1,141 @@
+'use strict';
+
+const Converter = require('./all-your-base');
+
+const converter = new Converter();
+
+describe('Converter', function () {
+
+  xit('single bit one to decimal', function () {
+    expect(converter.convert([1], 2, 10)).toEqual([1]);
+  });
+
+  xit('binary to single decimal', function () {
+    expect(converter.convert([1, 0, 1], 2, 10)).toEqual([5]);
+  });
+
+  xit('single decimal to binary', function () {
+    expect(converter.convert([5], 10, 2)).toEqual([1, 0, 1]);
+  });
+
+  xit('binary to multiple decimal', function () {
+    expect(converter.convert([1, 0, 1, 0, 1, 0], 2, 10)).toEqual([4, 2]);
+  });
+
+  xit('decimal to binary', function () {
+    expect(converter.convert([4, 2], 10, 2)).toEqual([1, 0, 1, 0, 1, 0]);
+  });
+
+  xit('trinary to hexadecimal', function () {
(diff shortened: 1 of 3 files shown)


exercises/all-your-base/all-your-base.spec.js
```diff
@@ -0,0 +1,140 @@
...
+
+const Converter = require('./all-your-base');
+
+const converter = new Converter();
+
+describe('Converter', function () {
+
+  it('wrong input_base base less than two', function () {
+    expect(function () {
+      converter.convert([0]);
```
rchavarria: To reflect better what line 9 describes, shouldn't this line be:

    converter.convert([0], 1);

Reading the test, it looks like `[0]` is the `inputBase` parameter, but in fact, it's `array` parameter.

What do you think? Do you think users will understand better this test adding the `inputBase` parameter as `1` (or another value less than `2`)?

exercises/all-your-base/all-your-base.spec.js
```diff
@@ -0,0 +1,140 @@
...
+
+  xit('wrong input_base base not integer', function () {
+    expect(function () {
+      converter.convert([0], 2.5);
+    }).toThrow(new Error('Wrong input base'));
+  });
+
+  xit('wrong output_base base less than two', function () {
+    expect(function () {
+      converter.convert([0], 2);
```
rchavarria: Same as comment on line 11

rchavarria:
Awesome!

It's nice to have a new exercise in this track.

After talking about comments on lines 11 and 23, would you mind to squash your commits? Would you need some help for doing that?

javaeeeee:
Hello Ruben,

Thank you for reviewing the code. I would like to mention that the tests
you commented on were intended to test the cases of missing bases as
JavaScript allows to pass as little arguments as you like. :)

There are separate tests that test unity as a base like 'first base is one'
and 'second base is one' on lines 99 and 105.

If you thank that testing the case of missing arguments is redundant, I can
remove them altogether. Or maybe, it's better to change the description of
the tests?

Thank you.

rchavarria:
You can keep tests without the parameter that are currently on L9 and L21. They show a specific feature of JavaScript. But, I would move the close to the ones you mentioned on L99 and L105 (I didn't notice them). And maybe, a better description could be `missing input/output base throws an error` or similar.

javaeeeee:
Hello Ruben,

Thank you for your comments. Renamed and moved the tests checking missing
inputs. Also squashed the commits.

rchavarria:
Thanks so much for this new exercise.

rchavarria:
I'm going to close this PR manually. I'm not sure why it wasn't marked as *merged* automatically when I pushed the merge :cry:

javaeeeee:
Thank you.

PR #314, opened 2016-12-26
dibs: I will implement exercise run-length-encoding

yonivy:


matthewmorgan:
@yonivy Thanks so much for doing this.  You've implemented all the canonical tests, and a nice clean example solution as well.

One request: you're missing a few semicolons, and the "house style" here is to include them everywhere they are normally required.  Can you fix that up, and then I would think we're ready to merge your contribution!

yonivy:
@matthewmorgan no problem. A few semicolons coming right up

yonivy:
@matthewmorgan do we have any "house style" linter config?

matthewmorgan:
@yonivy sadly, no!  This project isn't currently using linting at all.  I think that would be a good addition...know anyone who would be interested? ;) The [ecmascript track](https://github.com/exercism/xecmascript) uses it.

matthewmorgan:
Speaking of house style,  we usually like to have one commit per merged PR.  I should have squashed your commits...my bad. 

Thanks again for your contribution!

PR #320, opened 2017-01-25
Add 2 new test to check invalid input

aossama:
I've noticed that invalid inputs aren't handled well in the test cases, so I added mine, they are tested correctly in my exercism profile.


rchavarria:
Nice! It even makes this exercise more similar to [canonical tests](), and that is a desirable thing. Thanks.

Just one thing. It's common practice to merge PR with one single commit. Would you mind to squash them into a single one and update this PR?

If you find any problem squashig them, there is no problem, and the PR will be merged, so your contribution is highly appreciated.

aossama:
Sure thing will open another PR with one single commit.

Issue #321, opened 2017-01-27
Is unicode introduced too early in the list?


kytrinyx:
Bob is currently 5th exercise, and Word Count is currently the 7th exercise.

Both introduce unicode. If JavaScript is your first programming language, dealing with encoding is a pretty big hurdle. My guess is that a lot of people who tackle the JavaScript track on Exercism are new to programming.

For Bob it can be handled by the standard library function `toLocaleUpperCase()`, but if you're a beginner to programming, I'm not sure that you'd know to google for "string functions".

For Word Count it looks like we're tackling some pretty juicy stuff that isn't trivially solved with looking up string functions.

rchavarria:
Yep, I agree encoding is not the friendliest concept in programming. I don't remember exactly, but I think I solved one of them or both with regular expressions, and that's not very friendly neither for beginners.

I also think JavaScript is one of the first choices for beginners, because it's easy to start programming with it.

Are you suggesting we should move them further than the 10th position?

kytrinyx:
> Are you suggesting we should move them further than the 10th position?

Yeah. I think Bob is probably fine (I'd be curious to hear from some people who are doing the JavaScript track what they think, actually), but I would suggest moving Word Count out to position 20 at least.

kytrinyx:
I would recommend putting Pangram earlier, since it doesn't deal in unicode at all.

behrtam:
The unicode aspect was removed from many exercises in the `x-common` test data: https://github.com/exercism/x-common/issues/441

rchavarria:
Thanks for the reference @behrtam!

rchavarria:
As exercises were reordered, I'm going to close this issue.

PR #324, opened 2017-02-01
Move unicode problems up the list in `config.json`

rchavarria:
Word Count problem is moved to position 20th, because it introduces
users to a Unicode complex problem.

Bob problem, although likely dealing with Unicode subjects, is kept
because it can be reasonably solved without knowing too much about
Unicode.

Pangram is moved just before Bob.

Topics are configured in `config.json` for all three problems.

Addresses #321

### DIFF
--- a/config.json
+++ b/config.json
@@ -100,9 +100,28 @@
       ]
     },
     {
+      "slug": "pangram",
+      "difficulty": 1,
+      "topics": [
+        "Control flow (conditionals)",
+        "Control flow (loops)",
+        "Lists",
+        "Strings",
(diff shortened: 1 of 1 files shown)


rchavarria:
@exercism/javascript Nobody wants to review this? :cry:

rchavarria:
I feel lonely :stuck_out_tongue: 

If nobody on @exercism/javascript review this, I'll merge in on my own in ten days or so. All tests pass and it isn't a big change. But I want some feedback. A +1 would be enough to go on.

PR #326, opened 2017-02-17
Robot-simulator: require throwing exception

cltatman:
The original test passes without any assertions if `robot.orient` doesn't throw an exception

### DIFF
--- a/exercises/robot-simulator/example.js
+++ b/exercises/robot-simulator/example.js
@@ -10,6 +10,13 @@ Robot.prototype.at = function (xcoord, ycoord) {
 };
 
 Robot.prototype.orient = function (direction) {
+  if (direction != 'north' &&
+      direction != 'south' &&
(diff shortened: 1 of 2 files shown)


rchavarria:
Wonderful! Thanks for this contribution.

Could you modify the `/exercises/robot-simulator/example.js` file so that tests pass with your modification, please?

Before, tests didn't fail because the expectation was inside a `catch` block, but no `Error` was thrown, so the expectation was never executed (so the test passed in a wrong way).

Now, `example.js` will be more correct if you modify it to throw an `Error` when invalid parameters are used.

After that modification, it'd be awesome if you could squash all commits into a single one.

cltatman:
@rchavarria no problemo, example updated :+1:

rchavarria:
Absolutely awesome!!

Thank you.

PR #327, opened 2017-03-03
add more tests to drive towards complete solution

mattcbaker:
With the current test suite this submission will pass:

```
var PhoneNumber = function(raw){
    this.raw = raw
}

PhoneNumber.prototype.number = function() {
    return this.raw[0] === '2' || this.raw.length === 9 
        ? '0000000000' 
        : '1234567890'
}

PhoneNumber.prototype.areaCode = () => '123'

PhoneNumber.prototype.toString = () => '(123) 456-7890'

module.exports = PhoneNumber;
```

I've added some more tests to drive towards a more complete solution.  What does everyone think?

### DIFF
--- a/exercises/phone-number/phone-number.spec.js
+++ b/exercises/phone-number/phone-number.spec.js
@@ -11,13 +11,23 @@ describe('PhoneNumber()', function() {
     expect(phone.number()).toEqual('1234567890');
   });
 
+  xit('cleans some other number with dots', function() {
+    var phone = new PhoneNumber('555.456.7890');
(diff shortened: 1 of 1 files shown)


rchavarria:
Your new tests may seem redundant, but I think they're the perfect example of triangulation when practicing TDD to discover an algorithm. **Thanks for the contribution**.

In fact, you could consider to add some tests to the [canonical set of tests for phone-number](https://github.com/exercism/x-common/blob/master/exercises/phone-number/canonical-data.json), so that other language tracks can benefit of your contributions.

PR #332, opened 2017-03-20
Add exercise perfect-numbers

komyg:
Hi,

I would like to contribute implementing the [perfect-numbers](https://github.com/exercism/x-common/tree/master/exercises/perfect-numbers) exercise in Javascript.


matthewmorgan:
@komyg awesome.  When your PR is ready, please update the title to 'Add exercise perfect-numbers`.

Please let us know if we can help in any way.

komyg:
Hi @matthewmorgan,

I believe that I am mostly done with the implementation of this exercise, however TravisCI is complaining that it didn't found a module called Divisors, even though I included it into the divisors.js file. Is this a problem? Is there anything that I forgot to fix or configure?

Thanks,
Komyg

matthewmorgan:
@komyg yes, if you look at the output from Travis, it is complaining because you have not yet added the exercise to the `config.json` file in the root directory of the project.

See line 500: 
```
-> config.json does not include [perfect-numbers].
```

exercises/perfect-numbers/example.js
```diff
@@ -0,0 +1,40 @@
...
+    return 'Classification is only possible for natural numbers.';
+
+  // Factorize the current number.
+  var divsArray = divisors.getDivisors(number);
+
+  // Sum the factors.
+  sum = 0;
+  for (i = 0; i < divsArray.length; i++) {
+    sum = sum + divsArray[i];
+  }
```
matthewmorgan: Just a personal preference, but I thought I would mention the [array reduce](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/Reduce) method.

You could express this code like so:
```javascript
  // Sum the factors.
  sum = divsArray.reduce((sum, div) => sum += div, 0);
```

I mention that just for fun, not because you need to change it.

You could do something similar in your `Divisors` if you wished.

exercises/perfect-numbers/divisors.js
```diff
@@ -0,0 +1,33 @@
+'use strict';
+
+var Divisors = function() {
+
+}
```
matthewmorgan: need a semicolon here, and a few other places as well.  Let me know if you need me to call them out.

komyg: Sure, I will add the semicolons! Sorry, I didn't payed attention to them in the rest of the code...

exercises/perfect-numbers/divisors.js
```diff
@@ -0,0 +1,33 @@
...
+ * Calculate all the divisors for a given number and return them as an array.
+ * Note: the actual number is not include in the returned array.
+ */
+Divisors.prototype.getDivisors = function(number) {
+
+  var i;
+  var divs = new Array();
+
+  // Accepts only natura numbers greater than 1.
+  if (number <= 1)
```
komyg: I will add the curly braces as well.

exercises/perfect-numbers/example.js
komyg: Thanks for mentioning that. However this method uses ES6, right? With that arrow function `=>`. I thought that I was limited to ES5, because I tried to use the `import` keyword and the jasmine tests failed.

matthewmorgan: Good question.  NodeJS from 4 onward supports arrow functions, and 4 is what we recommend people install for this track.  4 is getting pretty stale, so it makes me think we should bump the version!  The `.yml` file specifies v5.7, so no worries about it building on Travis either.

If you prefer to use ES5 syntax, you can simply write:

```
sum = divsArray.reduce(function(sum, div) { return sum += div}, 0);
```

komyg:
Hi @matthewmorgan,

As for the divisors module, I thought my code would be more modular and easier to test if I added a separate function to calculate the divisors of a given number, however if I added this function to the example.js, then I would have to add the unit tests to the perfect-numbers.spec.js, which would in turn force the users to follow the same implementation that I did. So I created a new module in order to have a separate spec file.

I don't mind removing the divisors module and adding its function to the example.js directly. So what do you prefer? Should I keep it as the separate module (updating the config.json and make files), or should I add the divisors function to the example.js, but without a unit test?

matthewmorgan:
@komyg I suspected that's why you chose to use two modules, and kudos for both the clean design and the test coverage.

I agree that forcing the user into the same shape of implementation you chose is not appropriate. I think dropping the unit test is the pragmatic solution.

komyg:
All done! Please let me know if you need me to change anything else. :)

config.json
```diff
@@ -72,7 +72,8 @@
     "bowling",
     "diamond",
     "all-your-base",
-    "run-length-encoding"
+    "run-length-encoding",
+		"prefect-numbers"
```
matthewmorgan: Should be `perfect-numbers`.  How did this get past Travis?

matthewmorgan:
Looking very nice!  One more change please.

config.json
komyg: I have no idea... 

Maybe because I added the code below at the end of the config.json file.

```
   {
      "slug": "perfect-numbers",
      "difficulty": 1,
      "topics": [
      ]
    }
```
Anyway, I corrected the typo.

matthewmorgan:
@komyg looking great.  Maybe I was unclear on the last comment, but let's move this exercise up-- it doesn't belong at the end of the track.  Mind doing that please?  Then we can merge it!

>Also, let's move this higher up in the config.json, because it controls the order in which the exercises are served, and this one is less difficult than some at the end. How about right before word-count? This file is generally in need of reordering, so that's something we'll set up as chore, not a worry for this PR.

komyg:
Hey, I've changed the order of the `config.json` file as you requested. I am sorry, I didn't saw that last comment about it.

matthewmorgan:
@komyg Thanks!  Nice work.

komyg:
No, problem @matthewmorgan! I am happy to help! ;-)

matthewmorgan:
BTW @komyg if you are interested in porting this exercise to the EcmaScript track I would be happy to help.  It's not implemented there yet.  LMK https://github.com/exercism/xecmascript

PR #338, opened 2017-04-08
Edit leap year stub to be consistent with tests

jeffreydking04:
Rewriting the stub was trivial, but this is only the second exercise in the track and writing the class in this way makes complete sense, because we intuitively would want to instantiate a `Year` with a value, but it requires the beginning user to initialize the instance variable in one section of code and the `isLeap` logic in the instance method.  

Perhaps it would be better to leave the stub as it is and rewrite the tests so that `isLeap` takes the year parameter.  Even though this is a counter-intuitive design for the `Year` class, I believe it would be less intimidating for beginning users.  I would be glad to revise the tests if that is what is decided.

### DIFF
--- a/exercises/leap/HINT.md
+++ b/exercises/leap/HINT.md
@@ -1,9 +1,11 @@
 This is the first test for this exercise:
 
+```javascript
   it('is not very common', function() {
     var year = new Year(2015);
     expect(year.isLeap()).toBe(false);
   });
+```
 
 Each test in the exercise follows the same pattern: 
 
@@ -14,7 +16,9 @@ The `Year` function is a constructor, which means that it is specifically design
 
 When a new `Year` object is created:
 
+```javascript
   var year = new Year(2015);
+```
 
 We expect `year` to, at the very least, have a specific value (in this case 2015) assigned to it.  Otherwise it would not represent an actual year.
 
@@ -22,17 +26,21 @@ This means that we must store the value passed as a parameter when the new `Year
 
 The way a constructor stores these fundamental values (instance variables) is like this:
 
+```javascript
   var Constructor = function(input) {
     this.value = input;
   };
+```
 
 The `this` in the constructor refers to the newly created instance of the `Constructor`.
 
 Once the input (parameter) is stored in an instance variable, any instance methods, such as:
 
+```javascript
   Constructor.prototype.instanceMethod = function() {
     // this method can now access the input by calling `this.value`
   };
+```
 
 The instance method accesses the input using `this.value` (in this example).
 
--- a/exercises/leap/HINT.md
+++ b/exercises/leap/HINT.md
@@ -1,0 +1,44 @@
+This is the first test for this exercise:
+
+  it('is not very common', function() {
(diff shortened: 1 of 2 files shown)


matthewmorgan:
@jeffreydking04 with regard to:
> Perhaps it would be better to leave the stub as it is and rewrite the tests so that isLeap takes the year parameter.
we've had discussions like that on the Java track.  There are a couple of things to consider, which may be at some tension:

1. What makes sense to the beginner, especially this close to the beginning of the track?  You are correct to ask this question.
2. What promotes best practices in software design?

I would argue that (1) is potentially in conflict with (2), and that (2) should dictate the shape of the example code and the tests.  (1) could be better supported, if need be, by better documentation, perhaps a `HINT.md` file in this exercise.

The last thing to think about is the semantics.  Given the current semantics of instantiating a `Year`, I agree with you that the value of the year belongs in the constructor.  It doesn't make sense to have a `Year` object without that information.  So, the `isLeap()` check doesn't require a parameter.

If we were to move the parameter to `isLeap`, it would probably be best to change the class name to something like `LeapYearChecker` or some such.  Otherwise, we end up instantiating a `Year` that is...not really a Year.

@exercism/javascript @jeffreydking04 what are your thoughts?  I'm fine to merge this PR as is, but changing things up after a discussion would be OK too.  I would think we should choose between the two.

rchavarria:
I like the option of passing the year in the constructor. Thinking in OOP, it makes more sense.

I agree it's also a bit early to introduce these concepts in the second (or the first real) exercise, so it's a good idea to add a `HINT.md` file, to guide beginners to a solution.

The `HINT.md` file could show how to add a property to the `Year` object:

    function Year(aYear) {
        this.year = aYear;
    }

And then show them how to use that property in the method:

    Year.prototype.isLeap = function () {
         //  this.year; contains the value saved in the constructor
    }

jeffreydking04:
@matthewmorgan, @rchavarria: I agree that a `Year` class should be initiated with an actual year.  I will add a HINT.md to the PR for review soon.

jeffreydking04:
I added the HINT.md.  I figure that any user needing the hint would be struggling with constructors, instance variables, and instance methods in general, so I tried to communicate concisely those basic concepts.

matthewmorgan:
@jeffreydking04 this is a terrific contribution!  Very readable, enough information but not too much.

I wouldn't mind if @rchavarria would have a look, in case I overlooked something.  Otherwise, I think we can merge this.

matthewmorgan:
I realized one thing we will want to do, @jeffreydking04 , is to fix up the Markup formatting a bit.

For example, your code blocks can be formatted with triple-back-tick-javascript, so they look like:

```javascript
it('is not very common', function() { 
    var year = new Year(2015); expect(year.isLeap()).toBe(false); 
});
```

Would you mind making that change, and we'll see if @rchavarria has any other ideas for improvement?

Thanks!

rchavarria:
Thanks @jeffreydking04 and @matthewmorgan!! This is awesome!

PR #342, opened 2017-04-21
Update testing instructions on all exercises

joelwallis:
**Blocked by #344**

--

According to discussions in #341 we'll use Jasmine as a command line tool, globally installed through npm, and the testing command should now look like `jasmine ./test-file.spec.js`.

- [x] Change references for `jasmine-node` to `jasmine`  
  `❯ grep -lr "jasmine-node" .`
  - [x] `./docs/INSTALLATION.md`
  - [x] `./docs/TESTS.md`
  - [x] `./exercises/acronym/package.json`
  - [x] `./exercises/hello-world/HINTS.md`
  - [x] `./Makefile`
  - [x] `./package.json`
  - [x] `./SETUP.md`
- [x] Get #343 accepted
- [x] Get #344 accepted

PR #344, opened 2017-04-23
Robot Simulator example refactoring

joelwallis:
**Blocks #342**

--

The example for the Robot Simulator exercise [is not passing through unit tests](https://github.com/exercism/xjavascript/pull/342#issuecomment-296450529). The issue was already explained by Matthew [in his comment on #342](https://github.com/exercism/xjavascript/pull/342#issuecomment-296454271). I took the opportunity to make a complete refactoring, using a closure to encapsulate variables that are common to multiple methods or could be kept as singleton instead of an internal variable of an instance method.

The file is now passing all unit tests and nows seems to be easier to read and understand.

### DIFF
--- a/exercises/robot-simulator/example.js
+++ b/exercises/robot-simulator/example.js
@@ -1,88 +1,87 @@
-'use strict';
+var Robot = (function () {
+  'use strict';
 
-var Robot = function () {
-  this.coordinates = [0, 0];
-  this.bearing = 'north';
-};
-
-Robot.prototype.at = function (xcoord, ycoord) {
-  this.coordinates = [xcoord, ycoord];
-};
-
-Robot.prototype.orient = function (direction) {
-  if (direction != 'north' &&
-      direction != 'south' &&
-      direction != 'east' &&
-      direction != 'west') {
-    throw 'Invalid Robot Bearing'
+  var VALID_DIRECTIONS = ['north', 'east', 'south', 'west']
+  var INSTRUCTION_KEYS = {
+    A: 'advance',
+    L: 'turnLeft',
+    R: 'turnRight'
   }
 
-  this.bearing = direction;
-  return 'The robot is pointed ' + direction;
-};
+  function Robot () {
+    this.coordinates = [0, 0];
+    this.bearing = 'north';
+  };
 
-Robot.prototype.advance = function () {
+  Robot.prototype.at = function (x, y) {
+    this.coordinates = [x, y];
+  };
 
-  if (this.bearing === 'north') {
-    this.coordinates[1] += 1;
-  } else if (this.bearing === 'south') {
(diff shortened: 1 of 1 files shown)


matthewmorgan:
Nice work on the refactoring.  I especially appreciate that some of the names are clearer and that you replaced that chain of `if-else` with a `switch-case` block.

I was feeling like there might even be a less [cyclomatically complex](https://en.wikipedia.org/wiki/Cyclomatic_complexity) way to do that, like maybe with an object...and then I took a look back at my own solution to the problem, where I chose to do:

```javascript
RobotSimulator.prototype.advance = function(){
  var setCoords = {
    'north' : function(){this.coordinates[1]++},
    'south' : function(){this.coordinates[1]--},
    'east' : function(){this.coordinates[0]++},
    'west' : function(){this.coordinates[0]--}
  };
  setCoords[this.bearing].call(this);
};
```

Just thought I'd add that to the conversation.  I think the PR is just great as it is.

joelwallis:
Well, one thing that I just found weird is that, in Robot Simulator, when a robot facing north advances, its Y coordinate is increased. From what I understand of a two-dimension robot which faces north, he's actually "pointing up" on the screen - so, its Y coordinate should be decreasing instead of increasing.

joelwallis:
Well, I'll just create a new issue dedicated to discuss the north/south-related direction thing. :)

matthewmorgan:
I guess that depends on the frame of reference.  Must be whoever came up with this originally decided that 'up' > 'down'.  I think it's the same in all the language tracks I've looked at (which is a small sample to be sure :) )

joelwallis:
Yeah, but in order to take an object up on a 2D dimension you need to decrease the Y coordinate. CSS coordinates is a good example:

```css
#box {
  position: fixed;
  left: 50px;
  top: 50x;

  /* If I would like to move #box ten pixels up: */
  top: 40px;
}

matthewmorgan:
Yeah, I see your point.  Given that we're working in a JavaScript (IE, probably at least some front-end work) context, the choice of y-axis direction is at odds with the programming context.

My assumption, which I never examined until now, has always been that the coordinate system is based on the traditional two-dimensional [X-Y graph like in an algebra class](https://www.khanacademy.org/math/algebra-basics/alg-basics-graphing-lines-and-slope/alg-basics-coordinate-planee/a/points-on-the-coordinate-plane), where the origin is at the center and the first quadrant is positive in both coordinates.

Since the `README.md`, defined in [x-common](https://github.com/exercism/x-common/blob/master/exercises/robot-simulator/description.md), is pretty clear on the orientation, we would need to convince the community that it makes sense to use different conventions for different languages.

Does that make sense?  If you wanted to pursue changing this, I think the first thing to do would be to open an issue in [discussions](https://github.com/exercism/discussions).

PR #346, opened 2017-04-24
Adding proverb exercise

Erawpalassalg:
Adds the "proverb" exercise from the xcommon track to the xjavascript track.

Adds:
 - 1 folder "proverb" containing 2 files
    - example.js
    - proverb.spec.js

Changes:
  - config.json  (not sure where to place this exercise, therefore I put it in the last position)

If I missed something in order for the added exercise to work properly, let me know.

### DIFF
--- a/config.json
+++ b/config.json
@@ -48,6 +48,7 @@
     "series",
     "difference-of-squares",
     "secret-handshake",
+    "proverb",
     "linked-list",
     "wordy",
     "flatten-array",
@@ -350,6 +351,14 @@
       "slug": "secret-handshake",
       "difficulty": 1,
       "topics": [
+      ]
+    },
+    {
+      "slug": "proverb",
+      "difficulty": 1,
+      "topics": [
+        "Arrays",
+        "Optional values"
       ]
     },
     {
--- a/exercises/proverb/example.js
+++ b/exercises/proverb/example.js
@@ -1,0 +1,36 @@
+module.exports = function(){
+  var last = arguments[arguments.length - 1];
+  var chain = Array.from(arguments);
+  this.options = {};
+
+  if(typeof last == "object" && last.hasOwnProperty("qualifier")){
+    this.options = chain.pop();
+  }
+
+  this.chain = chain;
+  this.qualifier = this.options.qualifier ? this.options.qualifier + ' ' : '';
+
+  this.toString = function(){
+    return this.chainOfEvents() + this.conclusion();
+  }.bind(this);
+
+  this.chainOfEvents = function(){
+    return this.causesAndEffects().map( function(entry){
+      return "For want of a " + entry.cause +
+        " the "  + entry.effect + " was lost.\n";
+    }.bind(this) ).join('');
+  }.bind(this);
+
+  this.causesAndEffects = function(){
+     return this.chain.reduce( function(array, event, index){
+      if(index < this.chain.length - 1){
+        array.push({ cause: event, effect: this.chain[index + 1] });
+      }
+      return array;
+    }.bind(this), [] );
+  }.bind(this);
+
+  this.conclusion = function(){
+    return "And all for the want of a " + this.qualifier + this.chain[0] + ".";
+  }.bind(this);
+};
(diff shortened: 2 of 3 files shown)


config.json
```diff
@@ -507,6 +508,14 @@
       "difficulty": 1,
       "topics": [
       ]
+    },
+    {
+      "slug": "proverb",
+      "difficulty": 1,
+      "topics": [
```
rchavarria: In case you're not sure what topics include here, you can find a list of topics in https://github.com/exercism/x-common/blob/master/TOPICS.txt

config.json
```diff
@@ -507,6 +508,14 @@
...
+        "Array manipulation",
+        "arguments keyword"
```
rchavarria: I think this topic is too JavaScript specific. I would use some of the topics listed from the previous comment.

rchavarria:
Apart from comments on `config.json` about topics, I don't know about the position. Have you checked other tracks? What position do they use? It seems a similar exercise to Beer Song (11th position in this track). I'd place it 5 positions later Beer Song (between Grade School and Robot Name) unless it differs a lot compared to other tracks.

Another thing you could do is to squash the 3 commits into a single one. Let me know if you need help on this.

Erawpalassalg:
Hi,

Thanks for reviewing !

I squashed my commits (never done that before, since I used to see nuclear commits as a good practice but I can understand why they need to be squashed in that kind of project). Hope I did not break anything (it does not seem like I did).

From other tracks, the proverb exercise seems to be quite harder than the beer song. So I put it around the "wordy" and "secret-handshake" exercises, where it has been put in other tracks.

exercises/proverb/example.js
```diff
@@ -0,0 +1,35 @@
+module.exports = function(){
+  const last = arguments[arguments.length - 1];
```
matthewmorgan: We'll need to make this a `var` for the sake of consistency, and because this track emphasizes ES5.

exercises/proverb/proverb.spec.js
```diff
@@ -0,0 +1,78 @@
...
+
+    expect(proverb.toString()).toEqual(
+     "For want of a nail the shoe was lost.\n" +
+     "For want of a shoe the horse was lost.\n" +
+     "For want of a horse the rider was lost.\n" +
+     "And all for the want of a nail.");
+  });
+
+
+  it('tests Proverb class does not hard code the rhyme dictionnary',
```
matthewmorgan: Small typo here in `dictionary`.

exercises/proverb/example.js
Erawpalassalg: If the track emphasizes ES 5, I better change my arrow functions too, right ?

matthewmorgan: 😳  yep.  Good catch!

rchavarria: I saw the ES6 style in this file, but as it's *internal*, I said nothing. Anyway, I agree with @matthewmorgan, it'll be better to use *plain* ES5 as much as possible. Thanks!

Erawpalassalg:
Okay, 

The typo has been fixed and the code should now be all ES5. Thanks for your reviews and if I missed something, let me know. :)

matthewmorgan:
Awesome.  Just a few more things and I think we're ready to merge.  They're trivial, but I think it will help your `example.js` be as instructive as possible for users who dig into it for guidance:

1. You have included `index` in your `map()` call on line 18, but this argument is never used.  Let's omit it.
2. You don't need the semicolon on line 8.
3. There is a missing semicolon on line 36.

Re: the squash question, we do like to have everything for the PR squashed in to one commit, but GitHub gives us that option as part of the merge, so if you want to do it, feel free, otherwise we can do it on merge.

Erawpalassalg:
Done ! I also squashed the commits. Not sure it was the best thing to do, though, I think it gives less visibility during the review process. Anyway, you did not ask for much so it won't be a problem right now.

matthewmorgan:
:1st_place_medal: !

PR #347, opened 2017-04-25
Remove package.json and rely on global jasmine installation

tejasbubane:
I think this file is no longer needed.

@joelwallis forgot to remove it in https://github.com/exercism/xjavascript/pull/342#discussion_r112752492


joelwallis:
You're right!

But, well, since we'll be using a global installation of Jasmine, it makes sense to remote the `./package.json` file too.

tejasbubane:
It is being used by the [Makefile](https://github.com/exercism/xjavascript/blob/master/Makefile#L18). I guess we should change that too. We will then also need to change the [travis.yml](https://github.com/exercism/xjavascript/blob/master/.travis.yml#L9) to install jasmine first before testing. I will try in this PR itself. Thanks for bringing it up 👍

tejasbubane:
Done! I removed the `package.json` and made `Makefile` and `travis` rely on global `jasmine` installation. 😃

matthewmorgan:
Woohoo!!  Onward and upward.

PR #348, opened 2017-05-01
Add info about how to install Javascript

onealexharms:
So people don't have to go to stack overflow to get a link to that page like I did. :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,6 +2,10 @@
 
 Exercism exercises in JavaScript
 
+## Installing Javascript
+
+For information about installing Javascript, see the [Javascript](http://exercism.io/languages/javascript/about) page.
+
 ## Running Unit Test Suite
 
 The following commands assume that you are in the 'xjavascript' directory:
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@ Exercism exercises in JavaScript
 
 ## Installing
 
-To run the tests, you'll need node.js and Jasmine. For information on how to install them, see the [Javascript](http://exercism.io/languages/javascript/about) page.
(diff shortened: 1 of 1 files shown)


fredosantana:
JS is a markup language similar to HTML and CSS. No installation needed in general.

https://www.w3schools.com/js/

matthewmorgan:
@onealexharms yes this can use clarification for sure!  I think it might be worth explaining that JavaScript is interpreted by the V8 engine that is part of NodeJS, so installing Node gives us a JavaScript interpreter to run our tests and code.  

Would you mind adding that to your PR?

@fredosantana unless you install NodeJS, it would be difficult to run the test suites for this track.  Most modern OS's come with browsers installed, but I don't think they come with NodeJS installed, or any command-line JavaScript interpreter that I am aware of, so I think this PR has value.

onealexharms:
@matthewmorgan The page covers installing Node and npm, then Jasmine, so I think there's nothing to add. Let me know if you disagree?

matthewmorgan:
@onealexharms I agree that the page that you linked to has all the needed information around setting up the environment.  I was just wondering if a little context ahead of the link might be helpful, like:

> You'll need a JavaScript interpreter to run the test suites.  This comes bundled with NodeJS.  For information about how to install these tools, see...

What do you think?

onealexharms:
Oh, I actually like your wording better, but I didn't look at it when I typed... Should I change it, do you think?

matthewmorgan:
Sure, why not!  As long as we're changing it, let's do that.  I'm glad you brought our attention to this, I don't think we'd even noticed it needed touching up.

matthewmorgan:
:100:  Thanks!

Issue #313, opened 2016-12-16
Verify contents and format of track documentation


kytrinyx:
Each language track has documentation in the `docs/` directory, which gets included on the site
on each track-specific set of pages under [/languages](http://exercism.io/languages).

We've added some general guidelines about how we'd like the track to be documented in exercism/exercism.io#3315
which can be found at https://github.com/exercism/exercism.io/blob/master/docs/writing-track-documentation.md

Please take a moment to look through the documentation about documentation, and make sure that
the track is following these guidelines. Pay particularly close attention to **how to use images**
in the markdown files.

Lastly, if you find that the guidelines are confusing or missing important details, then a pull request
would be greatly appreciated.

rchavarria:
Mandatory files are present, and they explain what they're supposed to explain. But links to images should be fixed:

- [ ] On `TESTS.md`, link to *Solution explorer* should be fixed as it is recommended on exercism.io documentation, with a relative path instead of `http://x.exercism.io/v3/tracks/javascript/docs/img/SolutionExplorer.png`
- [ ] On `TESTS.md`, link to *Add files* should be fixed as it is recommended on exercism.io documentation, with a relative path instead of `http://x.exercism.io/v3/tracks/javascript/docs/img/AddFiles.png`

@exercism/javascript, any other suggestion?

matthewmorgan:
@rchavarria I haven't had a chance to read this yet, but I will get to it by the weekend.  I will also check it out for ES.  Thanks!

rchavarria:
I label this as "good first patch". The goal is to fix some links on `/docs/TESTS.md` file, as described before in this issue (https://github.com/exercism/xjavascript/issues/313#issuecomment-268087800).

morrme:
@rchavarria @matthewmorgan PR submitted .

matthewmorgan:
Closing as wontfix per #556.

PR #352, opened 2017-05-07
TESTS.md: make image links relative

morrme:
See #313

### DIFF
--- a/docs/TESTS.md
+++ b/docs/TESTS.md
@@ -22,7 +22,7 @@ Install [Visual Studio Express 2013 for web](http://www.visualstudio.com/en-us/p
 
 To get started, you can download a Visual Studio solution that is already set up to work with JavaScript and the other languages that Visual Studio supports.
 
-![Solution Explorer](http://x.exercism.io/v3/tracks/javascript/docs/img/SolutionExplorer.png)
+![Solution Explorer](./img/SolutionExplorer.png)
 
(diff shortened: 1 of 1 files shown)


matthewmorgan:
@morrme Thanks for doing this work.  However, I think the relative pattern that you want in this case is not `./some-file-name.png` but rather `/some-file-name.png` (no dot.)  

This is because the browser needs to resolve the file relative to the site root, where the dot syntax would be for resolving in a filesystem.

I don't have the exercism.io webapp installed locally, but I'm pretty confident you will need to make this change.  Let me know if you think otherwise.  I should be able to find some time to install the app and try your changes locally.

morrme:
When I delete the dot, the image does not render (the left is the current code, the broken link on the right is the result of the change): 
nodot

When I add the dot, it returns.

matthewmorgan:
OK, I think I should try to stand up the exercism UI app locally and see what happens.  Let me try this and I'll report my findings.

matthewmorgan:
OK, I'm having a bear of a time setting the exercism.io UI app up locally.  But, I found a solution that should work for _both_ the local filesystem and the web, EG:

```markdown
img/SolutionExplorer.png
```

No dot, no slash.  I believe this will work bc of the structure.  Shall we try that?

matthewmorgan:
Woohoo!  Thanks @morrme ! 🥇

rchavarria:
Thanks @morrme for this contribution!!

PR #358, opened 2017-05-20
Add the alphametics exercise

arthurchipdean:
This implements the alphametics exercise. Thanks

### DIFF
--- a/exercises/alphametics/alphametics.spec.js
+++ b/exercises/alphametics/alphametics.spec.js
@@ -1,0 +1,95 @@
+var solve = require('./alphametics');
+
+describe("Solve the alphametics puzzle", function() {
+
+  it("puzzle with three letters", function() {
+    var puzzle = "I + BB == ILL";
+    var expected = {
+      "I": 1,
+      "B": 9,
+      "L": 0
+    };
+    expect(solve(puzzle)).toEqual(expected);
+  });
+
+  xit("solution must have unique value for each letter", function() {
+    var puzzle = "A == B";
+    expect(solve(puzzle)).toBeNull();
+  });
+
+  xit("leading zero solution is invalid", function() {
+    var puzzle = "ACA + DD == BD";
+    expect(solve(puzzle)).toBeNull();
+  });
+
+  xit("puzzle with four letters", function() {
+    var puzzle = "AS + A == MOM";
+    var expected =  {
+      "A": 9,
+      "S": 2,
+      "M": 1,
+      "O": 0
+    };
+    expect(solve(puzzle)).toEqual(expected);
+  });
+
+  xit("puzzle with six letters", function() {
+    var puzzle = "NO + NO + TOO == LATE";
+    var expected =  {
+      "N": 7,
+      "O": 4,
+      "T": 9,
+      "L": 1,
+      "A": 0,
+      "E": 2
+    };
+    expect(solve(puzzle)).toEqual(expected);
+  });
+
+  xit("puzzle with seven letters", function() {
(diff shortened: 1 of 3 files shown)


exercises/alphametics/example.js
```diff
@@ -0,0 +1,107 @@
...
+
+function assignNumbers(numbers, letters, orders) {
+  var output = {};
+  for(var i = 0; i < letters.length; i++) {
+    output[letters[i]] = numbers[orders[i]];
+  }
+  return output;
+}
+
+function getUniqueLetters(string) {
```
matthewmorgan: Interesting!  I have never thought of using `indexOf` this way.  In ES6 we have `Set` now, which cleans up the semantics a little...

exercises/alphametics/example.js
```diff
@@ -0,0 +1,107 @@
...
+  }
+  return permute(inputArr);
+}
+
+function getNumberCombinations(set, k) {
+  var i, j, combs, head, tailcombs;
+  if (k > set.length || k <= 0) {
+    return [];
+  }
+  if (k == 1) {
```
matthewmorgan: Let's use `===` here.

exercises/alphametics/alphametics.spec.js
```diff
@@ -0,0 +1,89 @@
...
+  it("puzzle with three letters", function() {
+    var puzzle = "I + BB == ILL";
+    var expected = {
+      "I": 1,
+      "B": 9,
+      "L": 0
+    };
+    expect(solve(puzzle)).toEqual(expected);
+  });
+  xit("solution must have unique value for each letter", function() {
```
matthewmorgan: Would you mind putting a blank line between each `it` block please?

exercises/alphametics/alphametics.spec.js
```diff
@@ -0,0 +1,89 @@
...
+    var puzzle = "A == B";
+    var expected = null;
+    expect(solve(puzzle)).toEqual(expected);
```
tejasbubane: I think we should make this `.toBeNull()` instead.

exercises/alphametics/example.js
arthurchipdean: I agree. I plan on converting this to ES6 soon, there are lots of rooms for optimizations.

arthurchipdean:
@matthewmorgan  I rebased my branch and got rid of the extra commits but it doesn't seem like it actually removed them from the PR. In fact it looks like it duplicated existing ones? What am I doing wrong?

matthewmorgan:
@arthurchipdean I don't usually use `rebase` to squash, it's really more power than you need for this use case.  If you wanted to pick _some_ commits in the history, squash others, then `rebase` is the way to go.  Otherwise, I would just (make sure you're on your branch):

```bash
//this should squash all the commits you made on your branch down to one single one that follows the final commit on master
git reset --soft hash-of-last-commit-on-master
git commit -am 'Add the alphametics exercise'

//then force push back to your branch on GitHub:

git push -f origin master
```

The above should do it.

arthurchipdean:
@matthewmorgan Thank you!

arthurchipdean:
The tests are passing on my pc. I'm not sure what this error is: `gzip: stdin: not in gzip format` Is there something else I need to fix?

tejasbubane:
> The tests are passing on my pc. I'm not sure what this error is: gzip: stdin: not in gzip format Is there something else I need to fix?

@arthurchipdean That was some issue with out config validator. I restarted the build and its green now! 😄

matthewmorgan:
Thanks @arthurchipdean !

Issue #359, opened 2017-05-21
isogram: Add a new test with the same non-ASCII character in lower an upper case


rchavarria:
As discussed in #323, a new test could be added to the `isogram` exercise, one that test non-ASCII characters in lower an upper case, such as `Éléphant`

Zero-man:
I'd be happy to add this one. I'll open a PR soon.

Issue #323, opened 2017-01-30
isogram: suggested additional test - capitalized duplicate non-ASCII letter


kytrinyx:
Since the isogram exercise is dealing in non-ascii letters, I would suggest including a test that has a non-ASCII letter that is used in both uppercase and lowercase.

rchavarria:
Tests include words such as `Heizölrückstoßabdämpfung` and `éléphant`. Do they include non-ASCII letters?

I think they do (maybe I don't understand exactly what you're saying because of my English), so do you mean we could use words such as: `äbcdë ÄBCDË` or `Éléphant`?

And we could also update canonical tests, because they don't use non-ASCII letters. Well, I have some doubts here, I don't know the `json` format allows that.

kytrinyx:
Yeah, I meant "Éléphant", the same non-ASCII letter in both uppercase and lowercase.

I think we would need to have all the unicode stuff in a separate section of the canonical data so that tracks could choose to include it or not.

rchavarria:
I'd like to reopen this one, because I think nobody did something about it (I'm guilty in the first place)

Pending tasks:

- [ ] Add a new test with the same non-ASCII character in lower an upper case, such as `Éléphant`
- [ ] Add the same test to the canonical data, or in some other place so that people in other tracks can update this exercise

matthewmorgan:
@rchavarria maybe those could be called out as two separate issues since they're in two separate repos?  And both seem like a 'good first patch' to me.  What do you think?

rchavarria:
I added a new issue in this repo.

About adding it to x-common, I'm not sure. I've found this x-common issue https://github.com/exercism/x-common/issues/441 where people discussed about removing non-ASCII characters for this kind of exercise.

We could discuss here or in the new issue (#359) if we want to use non-ASCII chars or not.

PR #363, opened 2017-06-01
Add extended ascii character test for isogram.

Zero-man:
Tests a word which includes both upper and lower case of an extended ascii character. See issues [#359](https://github.com/exercism/xjavascript/issues/359) and [#323](https://github.com/exercism/xjavascript/issues/323).


rchavarria:
Many thanks for this contribution

PR #311, opened 2016-12-05
dibs: I will implement the simple-linked-list exercise

javaeeeee:
dibs: I will implement the simple-linked-list exercise

PR #372, opened 2017-06-26
Adds the simple-linked-list exercise (#311)

joelwallis:
I needed to implement this as a new branch since the work in progress of #311 was in the contributor's fork. This is my first exercise contribution, so please consider that I don't exactly know what I am doing 😂

Suggestions for improvements are more than welcome! 🎉

### DIFF
--- a/exercises/simple-linked-list/example.js
+++ b/exercises/simple-linked-list/example.js
@@ -1,0 +1,119 @@
+function Element (value, next) {
+  if (!(this instanceof Element)) {
+    throw new Error('Element is a constructor.');
+  }
+
+  if (value === undefined) {
+    throw new Error('Value required.');
+  }
+
+  if (next !== undefined && !(next instanceof Element)) {
+    throw new Error('A Element instance as next value is required.');
+  }
+
+  this.value = value;
+  this.next = next;
+};
+
+function List () {};
+
+List.prototype.push = function (value) {
+  if (value === undefined) {
+    throw new Error('Argument required.');
+  }
+
+  var newEl = (value instanceof Element)
+    ? value
+    : new Element(value);
+
+  if (!this.head) {
+    this.head = newEl;
+    return;
+  }
+
+  var lastEl = this.head;
+  while (lastEl.next) {
+    lastEl = lastEl.next;
+  }
+
+  lastEl.next = newEl;
+};
+
+List.prototype.unshift = function (value) {
+  if (value === undefined) {
+    throw new Error('Argument required.');
+  }
+
+  var newEl = (value instanceof Element)
+    ? value
+    : new Element(value);
+
+  newEl.next = this.head;
+  this.head = newEl;
+};
+
+List.prototype.shift = function () {
+  if (!this.head) {
+    return;
+  }
+
+  this.head = this.head.next;
+};
+
+List.prototype.pop = function () {
+  if (!this.head) {
+    return;
+  }
+
+  var penultEl, lastEl = this.head;
+  while (lastEl.next) {
+    penultEl = lastEl;
+    lastEl = lastEl.next;
+  }
+
+  if (!penultEl) {
+    this.head = undefined;
+  }
+  else {
+    penultEl.next = undefined;
+  }
+};
+
+List.prototype.reverse = function () {
+  if (!this.head) {
+    return;
+  }
+
+  var current, previous;
+  while (this.head) {
+    current = this.head;
+    this.shift();
+    current.next = previous;
+    previous = current;
+  }
+
+  this.head = previous;
+};
+
+List.prototype.toArray = function () {
+  var array = [];
+
+  while (this.head) {
+    array.push(this.head.value);
+    this.shift();
+  }
+
+  return array;
+};
+
+List.fromArray = function (array) {
+  var list = new List();
+  array.forEach(function (item) {
+    list.push(new Element(item));
+  });
+
+  return list;
+};
+
+exports.List = List;
+exports.Element = Element;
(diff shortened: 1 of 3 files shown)


joelwallis:
Forgot to add the exercise to `config.json`. Now tests are passing.

Also, I'm not sure which difficulty I should attribute to it, but I think `1` is not a fair one. A beginner would struggle to implement this one. Also, any suggestion for topics?

exercises/simple-linked-list/example.js
```diff
@@ -0,0 +1,72 @@
+function ListElement (value, next) {
+  if (!(this instanceof ListElement))
```
matthewmorgan: Let's use curly braces here, even though they're not strictly needed.  It's the style we've been using.
Same elsewhere-- I won't call out each spot.

exercises/simple-linked-list/example.js
```diff
@@ -0,0 +1,72 @@
...
+
+function LinkedList () {
+}
+
+LinkedList.prototype.push = function (value) {
+  if (value === undefined) throw new Error('Argument required.');
+
+  var newEl = (value instanceof ListElement)
+    ? value
+    : new ListElement(value)
```
matthewmorgan: Semicolon at the end of this line.

exercises/simple-linked-list/example.js
```diff
@@ -0,0 +1,72 @@
...
+  if (!this.head) {
+    this.head = newEl;
+    return;
+  }
+
+  var lastEl = this.head;
+  while (lastEl.next) lastEl = lastEl.next;
+
+  lastEl.next = newEl;
+}
```
matthewmorgan: Semicolon after here too.  Again, I won't call it out everywhere, but our current style is to use them by default.

exercises/simple-linked-list/simple-linked-list.spec.js
```diff
@@ -0,0 +1,129 @@
...
+var LinkedList  = SimpleLinkedList.LinkedList;
+var ListElement = SimpleLinkedList.ListElement;
+
+describe('simple-linked-list', function () {
+
+  it('exports a LinkedList constructor', function () {
+    expect(LinkedList).toBeDefined();
+  });
+
+  it('exports a ListElement constructor', function () {
```
matthewmorgan: don't forget to `xit` all the tests except the first one :)

exercises/simple-linked-list/example.js
```diff
@@ -0,0 +1,72 @@
+function ListElement (value, next) {
```
matthewmorgan: So, this is a nitpick, but the canonical README says:

> Write a simple linked list implementation that uses Elements and a List.

I think with that in mind, the classes should be changed to be called `Element` and `List`.  Since these aren't reserved words in JavaScript, we should be OK.

joelwallis: `Element` and `List` would be much better! Change is on its way

exercises/simple-linked-list/example.js
joelwallis: Roger that!

Also, where exactly did we set our coding standards? I remember we had a discussion about it but I can't remember if we decided to adopt a specific one.

exercises/simple-linked-list/example.js
joelwallis: Should I use semicolons to end statements. If that's the case, I'll add them across the file. I think I've coded it mostly following the JS Standard Style (no problem on changing this, just want to make sure).

joelwallis:
Well, that's weird. I've just pushed the requested changes but the GitHub Review doesn't reflect it. Probably because I used `--amend` while committing.

Anyway, it's available here: https://github.com/exercism/javascript/commit/59419edd54e59ce1cd03efcfb538c751962d5d20

joelwallis:
Hey @matthewmorgan!

I added tests for `List#reverse`, `List.toArray` and `List#toArray` methods, alongside an implementation example. Also, I changed the difficulty from 1 to 8. Let me know if there's anything else missing. 😊

exercises/simple-linked-list/example.js
```diff
@@ -0,0 +1,119 @@
...
+    return;
+  }
+
+  var lastEl = this.head;
+  while (lastEl.next) {
+    lastEl = lastEl.next;
+  }
+
+  lastEl.next = newEl;
+}
```
matthewmorgan: I hate to even mention this on your beautiful pull request, but all these function expressions should have terminating semicolons...this is why we need a linter!  Would you mind adding them in the meantime?

matthewmorgan:
This looks terrific.  Nice work sir!

If you wouldn't mind adding some more semicolons in a couple places...this is otherwise ready to merge!

exercises/simple-linked-list/example.js
joelwallis: Hey @matthewmorgan, no need feel bad about it. Regarding the function expression, you meant all function declarations? Named and unnamed ones? E.g.:

```js
// named function
function foo () {}
// anonymous function
var foo = function () {}
```

So, both of them should end with semicolons, right?

```js
// named function
function foo () {};
// anonymous function
var foo = function () {};
```

joelwallis: Well, I just pushed this change, assuming that both named and anonymous should receive semicolons.

joelwallis:
@matthewmorgan done! I'll wait for the CI to complete the tests, then I'll merge this PR. Thank you very much for your review!

matthewmorgan:
@joelwallis Nice job!  Thanks for doing this.

rchavarria:
I'm too late to reply but here are my two cents: I agree on the difficulty level and the topic. The bad news is that the topic doesn't exist in the [topic list](https://github.com/exercism/problem-specifications/blob/master/TOPICS.txt) (is it official?).

Anyway, I think the better topic to describe the exercise is `Data structures`, so it would be a matter of propose it to be added to the list. If we want to add more topics, I would add: `Arrays`, `Lists`.

matthewmorgan:
@rchavarria thanks for catching that!  @joelwallis would you mind creating a follow-up issue for that?

joelwallis:
@matthewmorgan definitely not! @rchavarria thanks for catching! I'll PR it right now

Issue #373, opened 2017-06-26
Add linting to Makefile


matthewmorgan:
We shouldn't require reviewers to comment on things like whitespace, semicolons, etc that a linter can easily call out.

This story is _not_ to lint exercise code that users are working on.  It is for submissions to the `exercism/javascript` repo.

Success Criteria:
- ESLint is integrated into the development workflow either with an `.eslintrc` file, or the appropriate config in `package.json`
- code is linted as part of the `make test` task.

KamilMalek:
Not sure about it, but here is my proposal

matthewmorgan:
Awesome!  We have some prior work on the Ecmascript track that we can use for guidance here.  I've made more details comments on your PR.

This is an important change that will really help move the track forward.  Let me know how I can help further.

rchavarria:
#399 is still open, but as #411 has been merged, I think this issue can be closed.

PR #374, opened 2017-06-27
Add linting to Makefile

KamilMalek:
#373

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,0 +1,12 @@
+{
+  "rules": {
+    "indent": [1, 2],
+    "no-trailing-spaces": 1, 
+    "eol-last": [1,"unix"],
+    "semi": [1,"always"], 
+    "linebreak-style": [1,"unix"] 
+  },
+  "parserOptions": {
+    "ecmaVersion": 6
+  }
+}
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ node_js:
   - 5.7
 
 install:
-  - "npm install -g jasmine"
+  - "npm install -g jasmine eslint"
 
 script:
   - "make test"
--- a/Makefile
+++ b/Makefile
@@ -15,6 +15,7 @@ TSTFILE := "$(subst _,-,$(ASSIGNMENT)).spec.$(FILEEXT)"
 test-assignment:
 	@echo "running tests for: $(ASSIGNMENT)"
 	@cp big-integer.$(FILEEXT) $(OUTDIR)
+	@eslint --fix exercises/$(ASSIGNMENT)/{$(TSTFILE),$(EXAMPLE)}
 	@cp exercises/$(ASSIGNMENT)/$(TSTFILE) $(OUTDIR)
 	@cp exercises/$(ASSIGNMENT)/$(EXAMPLE) $(OUTDIR)/$(subst _,-,$(ASSIGNMENT)).$(FILEEXT)
 	#@sed -i.original 's/\bxit\b/it/g' $(OUTDIR)/*spec.$(FILEEXT)


matthewmorgan:
@KamilMalek thanks for taking this on.  This is a big change, and an important one.  

I think there are a few things that we'll need to do to make this PR successful:
1. Get the rules right.  In the Ecmascript track, we've already implemented linting after discussing the rules.  Though we expect that we'll want to make additional changes to the rules going forward, we have a set right now that we're using.  They look like this:
```json
  "eslintConfig": {
    "parserOptions": {
      "ecmaVersion": 6,
      "sourceType": "module"
    },
    "env": {
      "es6": true,
      "node": true,
      "jasmine": true
    },
    "extends": "eslint:recommended",
    "rules": {
      "indent": [ "off", 2 ],
      "block-scoped-var": "off",
      "radix": "off",
      "no-use-before-define": "off",
      "one-var": [ "off", "always" ],
      "quotes": [ "off", "single", { "avoidEscape": true } ],
      "semi": [ "off", "always" ],
      "semi-spacing": [ "off", { "before": false, "after": true } ],
      "no-whitespace-before-property": "off",
      "space-before-blocks": [ "off", "always" ],
      "space-before-function-paren": [ "off", { "anonymous": "always", "named": "never" } ],
      "keyword-spacing": [ "off", { "before": true, "after": true } ],
      "no-multi-spaces": "off",
      "no-trailing-spaces": "off",
      "curly": [ "error", "all" ],
      "brace-style": [ "error", "1tbs", { "allowSingleLine": true } ],
      "object-curly-spacing": [ "off", "always" ]
    }
  }
```

We need to set the environment (`node', `jasmine`) to allow some globals like `require`, `describe`, etc that we use all over the place.  

2. Lint and fix all the exercises and tests.  This is a big one, but not as big as it seems: a lot of the changes are taken care of by the rule set above with the `--fix` flag that you set.  Some things will need to be addressed manually, like incorrect (or missing) variable declarations.  

3. Change the docs to explain to contributors that they need to install ESlint, and that linting is part of the `make` process.

I'm **happy to help** with any or all of the above steps.  We could discuss it here, or I could pull your changes into a new branch and add more changes in a new PR.  

Let me know how you want to proceed on this one.

rchavarria:
The rules @matthewmorgan commented above are the current ones, but I'm working on a new set of rules for the `ecmascript` track ([link to my fork](https://github.com/rchavarria/ecmascript/blob/lint-airbnb/package.json), search for `eslintConfig`).

In fact, the idea would be not to discuss about every single rule. As @joelwallis pointed out (https://github.com/exercism/ecmascript/issues/298#issuecomment-305335959), it's kind of useless. So, we decided to give [airbnb style guide](https://github.com/airbnb/javascript) a try.

The configuration should be much easier. Fix the errors will be tedious anyway.

rchavarria:
About the content of this PR:

- In the `ecmascript` track, we configure ESLint in the `package.json` file, because we want to spread that configuration to all exercises. In this track, I agree on using `.eslintrc.json` file (note `.json` extension, I think it's not recommended to use `.eslintrc` without extension any more), because we only want the linter for development purposes.
- I think the `eslint` command in the `Makefile` should be an independent *task*, separated from `test` or `test-assignment`. That way, it would be enough to run one `eslint` command like `eslint --fix exercises/**`. I would like to have both specs and implementations linted.

matthewmorgan:
Ahh.. timely and helpful feedback from @rchavarria!  @KamilMalek do you mind putting this one on ice until we have the related issue in Ecmascript sorted?  I think then we'll have a pattern we can use.

Of course, in the meantime, please feel free to comment on that issue: https://github.com/exercism/ecmascript/issues/298

rootulp:
I agree with @rchavarria in conforming to a style guide rather than selectively picking ESLint rules. Can we resurrect this issue? I think linting this repo is a great idea.

rchavarria:
In the ecmascript track we chose the Airbnb style. We should use the same here.

That style needs a couple of Node.js dependencies. I'm wondering if we should add a `package.json` file in this repo. It could manage these dependencies at the repo level: jasmine, eslint, <airbnb-style-guide-dependencies>. I mean, it's ok to execute `npm install` if you're a track maintainer/contributor, but I wouldn't like to get exercism users to execute that. Users will write solutions and run tests as they're doing now.

@exercism/ecmascripts - any thought?

rootulp:
@rchavarria repo level `package.json` makes sense especially b/c that's how [exercism/ecmascript](https://github.com/exercism/ecmascript/blob/master/package.json) does it

tejasbubane:
> Users will write solutions and run tests as they're doing now.

That is convenient 👍 

> In the ecmascript track we chose the Airbnb style. We should use the same here.

Agreed. Since having the same linting rules across the two repos will make it consistent and easy - especially for maintainers.

rchavarria:
Ok, let's add a `package.json` to the repo.

Now, I don't know who will do it. @KamilMalek started this PR to add linting to the repo. On the other hand, @rootulp seems to be interested too.

@KamilMalek, are you still interested in this?

KamilMalek:
Don't know if I will have any time soon, so I can pass it on.

rootulp:
@rchavarria @KamilMalek Added in #411

rchavarria:
Closing this because #411 has been merged. Feel free to ask to reopen it if I'm wrong.

Anyway, thanks @KamilMalek for starting it!

PR #378, opened 2017-07-04
port list-ops exercise

vivshaw:
I ported list-ops. Currently I have all the canonical test implemented; I could add some more in-depth ones like [some of the other tracks have](https://github.com/exercism/csharp/blob/master/exercises/list-ops/ListOpsTest.cs) if you like.

Let me know if there's anything up with my formatting; couldn't find a style guide for this track.

### DIFF
--- a/exercises/list-ops/example.js
+++ b/exercises/list-ops/example.js
@@ -1,0 +1,76 @@
+'use strict';
+
+function List(arr) {
+  this.values = arr || [];
+}
+
+List.prototype = {
+  append: function (otherList) {
+    var appended = this.values;
+
+    for (var i = 0; i < otherList.length(); i++) {
+      appended.push(otherList.values[i]);
+    }
+
+    return new List(appended);
+  },
+
+  concat: function (otherList) {
+    return this.append(otherList);
+  },
+
+  cons: function (item, arr) {
+    var x = new List([item]);
+    var xs = new List(arr)
+    return x.append(xs).values;
+  },
+
+  foldl: function (func, start) {
+    var acc = start;
+
+    for (var i = 0; i < this.length(); i++) {
+      acc = func(this.values[i], acc);
+    }
+
+    return acc;
+  },
+
+  foldr: function (func, start) {
+    var acc = start;
+
+    for (var i = this.length() - 1; i >= 0; i--) {
+      acc = func(this.values[i], acc);
+    }
+
+    return acc;
+  },
+
+  length: function () {
+    var count = 0;
+    this.values.forEach(function () { count++; });
+
+    return count;
+  },
+
+  reverse: function () {
+    return new List(this.foldl(this.cons, []));
+  },
+
+  map: function (func, arr) {
+    var applyFuncThenCons = function (x, acc) {
+      return this.cons(func(x), acc);
+    }
+
+    return new List(this.foldr(applyFuncThenCons.bind(this), []));
+  },
+
+  filter: function (pred) {
+    var consIfPred = function (x, acc) {
+      return pred(x) ? this.cons(x, acc) : acc;
+    };
+
+    return new List(this.foldr(consIfPred.bind(this), []));
+  }
+}
+
+module.exports = List;
--- a/exercises/list-ops/list-ops.spec.js
+++ b/exercises/list-ops/list-ops.spec.js
@@ -1,0 +1,168 @@
+var List = require('./list-ops');
+
+describe('List', function () {
+  // predicates to filter by & functions to map
+  var isOdd = function (x) {
+    return x % 2 === 1;
+  }
+
+  var plusOne = function (x) {
+    return x + 1;
+  };
+
+  var divide = function (x, acc) {
+    return x / acc;
+  };
+
+  describe('append', function () {
+    it('appends two empty lists', function () {
+      var emptyList = new List();
+      expect(emptyList.append(emptyList).values).toEqual([]);
+    });
+
+    xit('appends an empty list to a non-empty list', function () {
+      var emptyList = new List();
+      var nonEmptyList = new List([1, 2, 3, 4]);
+      expect(emptyList.append(nonEmptyList).values).toEqual([1, 2, 3, 4]);
+    });
+
+    xit('appends two non-empty lists', function () {
+      var list1 = new List([1, 2]);
+      var list2 = new List([2, 3, 4, 5]);
+      expect(list1.append(list2).values).toEqual([1, 2, 2, 3, 4, 5]);
+    });
+  });
+
+  describe('concat', function () {
+    xit('concatenates an empty list', function () {
+      var emptyList = new List();
+      expect(emptyList.concat(emptyList).values).toEqual([]);
+    });
+
+    xit('concatenates a list of lists', function () {
+      var list1 = new List([1, 2]);
+      var list2 = new List([3]);
+      var list3 = new List([]);
+      var list4 = new List([4, 5, 6]);
+      expect(list1
+        .concat(list2)
+        .concat(list3)
+        .concat(list4).values
+      ).toEqual([1, 2, 3, 4, 5, 6]);
+    });
+  });
+
+  describe('filter', function () {
+    xit('filters an empty list by a function', function () {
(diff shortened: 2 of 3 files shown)


exercises/list-ops/example.js
```diff
@@ -0,0 +1,74 @@
+'use strict';
+
+function List() {
+  return {
+    append: function (arrOne, arrTwo) {
+      var appended = [];
+
+      for (var i = 0; i < arrOne.length; i++) {
```
matthewmorgan: Can we do this without relaying on `Array.length`?  I ask because I think the idea is to use as few built-in "list" methods as needed.  You should be able to get by with `push` and `pop` I believe.

exercises/list-ops/example.js
```diff
@@ -0,0 +1,74 @@
...
+      }
+
+      return appended;
+    },
+
+    concat: function (arr) {
+      var concatenated = [];
+
+      for (var i = 0; i < arr.length; i++) {
+        concatenated = this.append(concatenated, arr[i]);
```
matthewmorgan: Nice reuse of your own method here!

exercises/list-ops/example.js
```diff
@@ -0,0 +1,74 @@
...
+      }
+
+      return concatenated;
+    },
+
+    cons: function (x, arr) {
+      return this.append([x], arr);
```
matthewmorgan: And again, nice reuse.

exercises/list-ops/example.js
```diff
@@ -0,0 +1,74 @@
...
+      return acc;
+    },
+
+    length: function (arr) {
+      var increment = (x, acc) => acc + 1;
+      return this.foldl(increment, 0, arr);
+    },
+
+    reverse: function (arr) {
+      return this.foldl(this.cons.bind(this), [], arr);
```
matthewmorgan: Interesting!  I would not have thought of this approach.

exercises/list-ops/list-ops.spec.js
```diff
@@ -0,0 +1,95 @@
...
+      expect(list.concat([])).toEqual([]);
+    });
+
+    xit('concatenates a list of lists', function () {
+      expect(list.concat([[1, 2], [3], [], [4, 5, 6]])).toEqual([1, 2, 3, 4, 5, 6]);
+    });
+  });
+
+  describe('filter', function () {
+    var isOdd = x => x % 2 === 1;
```
matthewmorgan: Nice semantics.

vivshaw:
Thanks for the input!

1. Wow, I'm so used to building for loops with Array.length that I completely missed that! Should be easy enough to switch each use to push/pop or to use my own length() method. Will do!

2. Instance methods were my original intuition as well, but I got worried when all the other tracks I peeked at had used static methods & decided against it. Since you agree that instance methods make more sense here, I'll switch it back over to instances & see how things look.

exercises/list-ops/list-ops.spec.js
rchavarria: In this line (as well as `51` and `63`) makes use of fat arrow functions.

That's a language feature that's been released... 2 years ago? I'm not sure. Anyway, that language features are more appropriate for the `ecmascript` track. This track should stick to *old* ES5 JavaScript features.

It wouldn't  be a problem for the `example.js` file, but tests are going to be visible for users. They should not expect ES6 (or newer) features in exercises of this track.

It's not a problem with the code, it's a problem with the philosofy/spirit of `javascript` and `ecmascript` tracks.

exercises/list-ops/example.js
vivshaw: fold is magic ✨
my thought process: cons is the identity function for foldr, so by switching to foldl, we get the same list back, but reversed!

rchavarria:
:+1: to instance methods as well.

Please, @vivshaw, if you don't mind, could you squash all commits into a single one? That will create a beautiful git history for the repo. Let me know if you need any help with this.

exercises/list-ops/list-ops.spec.js
vivshaw: OK, will remove the arrow functions.

matthewmorgan:
@rchavarria good catch on the ES features...

@vivshaw I think it's interesting that the Java track has it implemented in a static way, since I know they have come to a policy of preferring instance methods as a way of encouraging best practices.  But, maybe it's just housekeeping they haven't gotten to yet...

vivshaw:
Alright, I've removed the ES features, moved to instance methods, removed all use of `Array.length` in the examples, and squashed to 1 commit.

I also added some spies to test whether native array functions like `Array.prototype.map` and so on have been called.

3 remaining questions:
1. Different tracks have interpreted `concat` totally differently. Tracks like CSharp have interpreted the spec to mean 'make a function that flattens a list of lists', whereas other tracks like ES have interpreted `concat` to be synonymous with `append`. I've brought my implementation in line with the ES track for now.

2. There doesn't seem to be any 100% accurate replacement for `Array.length`. I'm now using a counter variable with forEach, which is fine for dense arrays, but gives incorrect results for sparse arrays:

```
function length(arr) {
  var count = 0;
  arr.forEach(function () { count++; });
  return count;
}

var a = [1, 2, ,, 3];
console.log(a.length) // returns 5
console.log(length(a)) // returns 3!!
```

3. The test suite is a little less elegant now after adding spies. I tried putting the tests on the `Array.prototype` methods in a tidy little beforeAll block so that we can listen in as we're doing the normal test suite, but it turns out that putting spies on `Array.prototype.concat()` and `Array.prototype.reverse()` will fail, as Jasmine's internal use of array methods will trigger the spies. So we need to spy directly on the `values` of a particular `List`. Unless I'm mistaken, there's no way to do this in a `beforeAll` block because the lists are defined inside each test. So, we need to make an entire new suite  just for the spies that defines a single `List`, spies on its `values`, re-runs each of the methods in the spec, and then we can test if our spies have been called. It's not pretty, but it works correctly. If you have any suggestions on making this tidier I'd be happy to implement them.

matthewmorgan:
@vivshaw thanks for getting these changes made.  To answer your questions:

1. I think your handling of `concat` makes sense for two reasons: firstly, that's how JavaScript's `Array.concat()` works, and secondly, this track already has a separate exercise `flatten-array`.  I don't think flattening is the intention here.
2. I think your `Array.forEach` approach is fine.  The `config.json` doesn't specifically address sparse arrays, and for this purpose I don't think we need to cover that.  But, in case it's of any interest to you: 
In ES5 you should be able to use `for...in` to get only the populated indices of an Array.  Like so:

```JavaScript
for (var el in a){ 
  console.log(a[el]);
}

//output 
1
2
3
```

Another possible approach would be filter using `Object.hasOwnProperty()`, returning only populated indices:

```JavaScript
a.filter((el, idx) function {
   a.hasOwnProperty(idx));
}

// output

[1, 2, 3]
```
3.  I think you've gone above and beyond with the spies!  My original feedback was just advice to you as the exercise author, not really to enforce the "no Array" constraint on the user.  However, I think since you've put it at the end of the test suite, I don't see any reason it can't be there as a final challenge for the users.  It's not hard to imagine someone implementing a `List` class that truly just wraps the native methods, and your addition would call that out.  @rchavarria how do you feel about that?

vivshaw:
@matthewmorgan thanks for pointing out the `for ... in` and `hasOwnProperty` approaches! I keep forgetting that JS arrays are also full-fledged `Objects` on which we can do the usual `Object` alchemy when needed. I've gotten too used to C++ & Java, I suppose.

And, if it's too much, I'm just as happy to take out the spies section as well; I just figured since Jasmine gives us the ability to easily test whether users are actually implementing the algorithms themselves or not that I'd throw that in.

matthewmorgan:
Thanks @vivshaw !  Please come back and contribute any time!

rchavarria:
My opinion about spies is that they could confuse users. They are advanced topics in testing, and users might find them difficult to understand.

But, the work is done, and we don't know if they'll work or not. So, let's keep them and we'll see if users find them useful or not.

In any case, many thanks @vivshaw for the contribution (and @matthewmorgan for such incredible feedback)

Issue #382, opened 2017-07-17
Nextercism configuration


rchavarria:
As stated in #381, Exercism will be moving to a tree-shaped track rather than a linear one, as described in the [progression & learning in Exercism](https://github.com/exercism/docs/blob/master/about/conception/progression.md) design document.

So, some tasks must be done: 

- [x] Add a rough estimate of difficulty to each exercise (scale: 1-10) if it hasn't already one
- [x] Add topics to exercises that has no topics yet
- [x] Choose *at most 20 exercises* to be core exercises (set `core: true`, and delete the `unlocked_by` key)
- [x] For each exercise that is not core, decide which exercise is the prerequisite (max 1)
- [x] If possible, leave 3 or 4 simple exercises as (`core: false`, `unlocked_by: null`), as this will provide new participants with some exercises that they can tackle even if they have not finished the first core exercise.

rchavarria:
I started working on branch [nextercism-config](https://github.com/exercism/javascript/tree/nextercism-config).

I plan to push commits as soon as they're ready, so that anyone can work on this branch too.

rchavarria:
Following @matthewmorgan's advice, copying Java guys, I've created a [spreadsheet](https://docs.google.com/spreadsheets/d/1pELRlzUlCSea5ybbBslFgoeLgDOkRZUjqZxN4ch2_1o/edit?usp=sharing) with current JavaScript exercises, I chosed 19 core exercises and unlocking exercises for non-core ones.

Of course, the spreadsheet shouldn't be the truth, so feel free to comment or edit it.

When everything looks more or less decided, I'll change `config.json` to reflect changes.

There aren't 3 or 4 exercises with `core: false`, `unlocked_by: null`. I think these exercises coudl be good candidates: queen-attack (difficult), pig-lating (easy), minesweeper (game, difficult), roman numerals (very well known).

Spreadsheet link: https://docs.google.com/spreadsheets/d/1pELRlzUlCSea5ybbBslFgoeLgDOkRZUjqZxN4ch2_1o/edit?usp=sharing

matthewmorgan:
@rchavarria thanks for doing this.

A question: according to your task list above, the `core: false` exercises should be "simple." But, I don't think queen-attack or minesweeper would qualify as simple in the context of this track.  IS the requirement to use simple exercises or any exercises?

rchavarria:
[...]

PR #383, opened 2017-07-18
Add topics and difficulty to all exercises

rchavarria:
This PR acomplishes the first two tasks of #382 

- Add a rough estimate of difficulty to each exercise (scale: 1-10) if it hasn't already one
- Add topics to exercises that has no topics yet

Of course, topics might be wrong (almost all exercises are about conditionals and loops, but that's programming), difficulty might be wrong. But we can discuss now that we have some values for them.

Next step: Choose *at most 20 exercises* to be *core* exercises (set `core: true`, and delete the `unlocked_by` key)

### DIFF
--- a/config.json
+++ b/config.json
@@ -35,7 +35,7 @@
       "slug": "hamming",
       "core": false,
       "unlocked_by": null,
-      "difficulty": 1,
+      "difficulty": 2,
       "topics": [
         "Control-flow (loops)",
         "Control-flow (conditionals)",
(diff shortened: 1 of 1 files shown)


matthewmorgan:
Nice work Rubén!  This is a very big task.  The guys on the Java track approached fine-tuning the difficulties by creating a spreadsheet and all the maintainers entered their opinion of the difficulty of each exercise.  That's something we could do to compare notes for the next refinement, but I think this is a great start.

rchavarria:
That sounds good. We could do that once we finish all tasks of #382

PR #385, opened 2017-07-23
Add exercise sublist

matthewmorgan:
A port of the EcmaScript version.

BTW re-pushed this bc I forgot to update the README for JS testing...

### DIFF
--- a/config.json
+++ b/config.json
@@ -189,6 +189,17 @@
       ]
     },
     {
+      "uuid": "4a83a72c-db0a-45b6-b77c-1949cb24fbae",
+      "slug": "sublist",
+      "difficulty": 4,
+      "core": false,
+      "unlocked_by": "linked-list",
+      "topics": [
+        "Lists",
+        "Arrays"
+      ]
+    },
+    {
       "uuid": "b668e11a-a8ce-4e94-ba68-3a1f0fa3f6c8",
       "slug": "space-age",
       "core": false,
@@ -310,7 +321,7 @@
       "uuid": "f1943e87-182a-44f5-a885-3d68a0c0a0dc",
       "slug": "circular-buffer",
       "core": false,
-      "unlocked_by": null,
+      "unlocked_by": "linked-list",
       "difficulty": 8,
       "topics": [
         "Control flow (conditionals)",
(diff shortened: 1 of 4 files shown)


config.json
```diff
@@ -189,6 +189,16 @@
       ]
     },
     {
+      "uuid": "4a83a72c-db0a-45b6-b77c-1949cb24fbae",
+      "slug": "sublist",
+      "difficulty": 4,
+      "core": false,
+      "unlocked_by": null,
```
rchavarria: We haven't decided yet about what exercises unlock what exercises, but in this case I think we could set `unlocked_by` property.

`sublist` is an exercises that makes you implement some functionality that is provided by some languages or libraries, and it's related to lists. Do you think could it be unlocked by `linked-list` exercise?

I think `linked-list` could unlock: circular-buffer, binary-search, custom-set and simple-linked-list

config.json
```diff
@@ -189,6 +189,16 @@
...
+      "topics": [
```
rchavarria: Another topic could be `Arrays`.

I'm not doing a clear distinction between `Lists` and `Arrays` in JavaScript. For me, it's more or less like: `Arrays` is for the language itself (`[]`, `[1, 2, 3]`) and `Lists` is more about the programming concept, the data structure, manipulation of `Arrays` (in case of JS). As I said, it's not clear enough for me, so I'm not sure this explanation is clear for others, sorry.

exercises/sublist/sublist.spec.js
```diff
@@ -0,0 +1,126 @@
+var List = require('./sublist');
+
+
+describe('sublist', function() {
+
+  it('two empty lists are equal', function() {
+    const listOne = new List();
```
rchavarria: What about using `var` instead?

In this case, I don't have a strong opinion. We should try to use just ES5 code in this track. But, `const` is not a very advanced, or a very complex ES6 feature. In fact, it's a good first step towards ES6.

If it was a destructuring, I's strongly recomend going back to ES5. But it's a `const`, not a big deal.

What do you think? Should we stick to ES5 even in this case?

matthewmorgan: Yep!  This was an oversight on my part-- thanks for catching it!

matthewmorgan:
OK @rchavarria thank you for the suggestions.  Let me know what you think!

rchavarria:
Awesome, let's wait a little bit just in case any @exercism/ecmascripts wants to add some comments.

But, definitely, it should be merged before modifying `config.json` with core/non-core and so on.

rchavarria:
Let's merge this and start working at #382

PR #392, opened 2017-08-06
Fix typo

snoozer05:
add missing 'd' to obscured string


rchavarria:
Nice catch!

Tested with the current implementation and @snoozer05 is right, there's a missing `d`

PR #393, opened 2017-08-06
Add a new test to circular buffer exercise

netonion:
I've seen people came up with solution like
`Buffer.buffer = [];
Buffer.read = function() { /* do something with Buffer.buffer */ };
module.exports = Buffer;
`.
While it's not required in the description that multiple circular buffers can be instantiated, I felt that this kind of pattern should be discouraged, and there should be a test to check this.

### DIFF
--- a/exercises/circular-buffer/circular-buffer.spec.js
+++ b/exercises/circular-buffer/circular-buffer.spec.js
@@ -109,4 +109,12 @@ describe('CircularBuffer', function() {
     expect(buffer.read).toThrow(bufferEmptyException());
   });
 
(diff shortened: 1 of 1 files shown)


rchavarria:
Well, that pattern might be useful in some cases.

In this case, I see your point and the test makes complete sense for me.

@netonion - you might want to create a issue or propose a PR on [`problem-specifications`](https://github.com/exercism/problem-specifications) repo, to include this case as a canonical test case for the exercise (see [`canonical-data.json`](https://github.com/exercism/problem-specifications/tree/master/exercises/circular-buffer/canonical-data.json)) so that other tracks will benefit from the new test case.

PR #394, opened 2017-08-08
Add change exercise

Radix16:
I added a 'Change' exercise to the set Javascript exercises.   I based the Jasmine tests upon the set of tests used from the same exercise in the Java project.

### DIFF
--- a/config.json
+++ b/config.json
@@ -1032,6 +1032,14 @@
       ]
     },
     {
+       "uuid" : "910fe904-7e3c-11e7-bb31-be2e44b06b34",
+       "slug" : "change",
+       "core" :  false,
+       "unlocked_by": "prime-factors"
+       "difficulty" : 8,
+       "topics": ["Algorithms", "Mathematics", "Performance", "Searching"]
+    },
+    {
       "uuid": "1b53340d-ea40-44ee-bf2e-42e516704e7c",
       "slug": "nucleotide-count",
       "deprecated": true
--- a/exercises/change/change.spec.js
+++ b/exercises/change/change.spec.js
@@ -1,0 +1,73 @@
+
+
+
+var  Change =require('./change');
+
+describe('Change', function() {
+
+  it('change for 1 cent',function() {
+     var change= new Change();
+     var result = change.calculate([1,5, 10, 25],1);
+     expect(result).toEqual([1]);
+  });
+
+  xit('return a single coin',function() {
+    var change= new Change();
+    var result = change.calculate([1,5, 10, 25, 100],25);
+    expect(result).toEqual([25]);
+  });
+
+  xit('multiple coins coin',function() {
+    var change= new Change();
+    var result = change.calculate([1,5, 10, 25, 100],15);
+    expect(result).toEqual([5, 10]);
+  });
+
+  xit('test with Lillipution Currency where a greedy algorithm fails',function() {
+    // https://en.wikipedia.org/wiki/Change-making_problem#Greedy_method
+     var change= new Change();
+     var result = change.calculate([1, 4, 15, 20, 50],23);
+     expect(result).toEqual([4,4,15]);
+  });
+
+  xit('test with lower Elbonian Currency where a greedy algorithm fails',function() {
+    // https://en.wikipedia.org/wiki/Change-making_problem#Greedy_method
+    var change= new Change();
+    var result = change.calculate([1, 5, 10, 21, 25],63);
+    expect(result).toEqual([21, 21, 21]);
+  });
+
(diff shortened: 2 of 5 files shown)


Radix16:
The example solution was not in the branch when the CI process ran. I will re-submit a new PR

config.json
```diff
@@ -1031,6 +1031,11 @@
         "Lists"
       ]
     },
+      {
```
rchavarria: This object needs some more properties:
- `uuid`: sorry, I don't know where these ids are coming. What about copying the `uuid` from the Java track?
- `core`: it should be set to `false`, because we have already decided all *core* exercises
- `unlocked_by`: it should point to a core exercise. I'd suggest `leap`, `space-age` or `prime-factors`. Just choose one.

I'd also add some more topics, `Mathematics`, `Integers` could be some of them. You'll find here a list of [available topics](https://github.com/exercism/problem-specifications/blob/master/TOPICS.txt)

exercises/change/change.spec.js
```diff
@@ -0,0 +1,78 @@
+
```
rchavarria: You can check if you're missing some *canonical* test comparing this specs with [this json file](https://github.com/exercism/problem-specifications/blob/master/exercises/change/canonical-data.json).

rchavarria:
Sorry, I requested changes before making this comment. First of all I should thank you for your effort. It's awesome you want to contribute. Many, many thanks.

exercises/change/change.spec.js
Radix16: All of these tests are present in the pull request.  The pull request actually includes one additional, but trivial test.   The canonical tests return a -1 value for an error condition.  The pull request tests throw an error object, which is the same practice in the Java language tests.   Please advise if returning a -1 value is preferred.

config.json
Radix16: I added the uuid, core and unlocked_by values as requested.  I also added additional topics.   By the way,  I came across this issue for the exercise configlet https://github.com/exercism/configlet/issues/30.   According to this issue, the uuid values are supposed to be globally unique.   I generated a version 1 uuid from uuidgenerator.net and applied this value instead of copying the value from the Java track

Radix16:
Thanks for the feedback @rchavarria.   I have made changes that you requested.  Please let me know if there is anything else I need to correct.   

Working on this exercise has been more fun than I expected.   : )

exercises/change/change.spec.js
rchavarria: I'm glad all canonical cases are include, great!

I usually like more throwing errors than returning error codes in these exercises, so I'm ok with them too.

In this particular file, `change.spec.js`, I still see some indentation issues. Sorry to be so insistent with it. `xit` blocks should look like:

```javascript
  xit('test description', function () {  // indented two spaces compare to describe
    var change = new Change();           // indented two more spaces
    // ...
    expect(result).toEqual(...);
  });                                     // indented two spaces compare to describe
```

matthewmorgan:
@rchavarria could @Radix16 run `eslint --fix .` to address these indentation issues automatically?

Radix16:
@rchavarria I fixed the space in the change.spec.js file.   Please let me know if there are other changes to make.  Thanks

rchavarria:
Awesome @Radix16, it looks much nicer now!!

Would you mind to squash all commits of this PR into a single one? These commands may help you:

```
git reset --soft 55e4a38fb333d33b605c74678448c93769abd69a
git commit -am 'Add exercise change'
git push -f
```

If that doesn't help you, drop a line here and we'll be glad to help.

Radix16:
Hi @rchavarria.   I have squashed the commits as you requested.   Thanks for the command reference.  I am new to git.

Radix16:
@rchavarria @matthewmorgan is there anything else that I can do for this pull request?   There is a conflict on this patch from a more recent commit.   What do I need to do to solve?    Thanks

matthewmorgan:
@Radix16 I think it is ready to merge-- sorry that we have not gotten back with you sooner.

As far as resolving the conflict, there has been a change to the `config.json` format, so there are a few steps, but all of them are pretty simple:
1. Pull `exercism/javascript:master` in to your fork's master branch. (NB You may need to manually edit your `config.json` entry to match the new format.)
2. Rebase your changes on to master.
3. Push your changes back up to your branch.  This may require a `-f` for force push.

Some of these steps can be a little tricky if you're not familiar with them.  Please don't hesitate to ask for help, I am happy to assist you.  :)

matthewmorgan:
@Radix16 Looks like you have an additional commit now, but I will squash on merge for you.

Thanks for this addition!

PR #400, opened 2017-08-30
Avoid mutations on List#toArray() (simple-linked-list exercise)

joelwallis:
@tejasbubane caught this on exercism/ecmascript#345, while reviewing @apapirovski's ECMAScript version of this exercise

### DIFF
--- a/exercises/simple-linked-list/example.js
+++ b/exercises/simple-linked-list/example.js
@@ -97,10 +97,11 @@ List.prototype.reverse = function () {
 
 List.prototype.toArray = function () {
   var array = [];
+  var current = this.head;
 
-  while (this.head) {
-    array.push(this.head.value);
-    this.shift();
+  while (current) {
+    array.push(current.value);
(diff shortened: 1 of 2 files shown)


joelwallis:
Wow, 400th PR! 😲

joelwallis:
I added expectations to the `List#toArray()` test case, but I think it hurts the isolation of the test case itself. Maybe it would be better to create a separated test case for this scenario. Does it make sense or I'm just byte brushing here?

tejasbubane:
> but I think it hurts the isolation of the test case itself. Maybe it would be better to create a separated test case for this scenario. Does it make sense or I'm just byte brushing here?

We already have multiple `expect`s in a single test example, so I guess its fine.

matthewmorgan:
I agree with @tejasbubane. Especially because those "multi-expect" tests are working together to test the behavior that the test describes.

Thanks @joelwallis !

PR #404, opened 2017-09-04
ran new configlet fmt as told in the documentation

Vankog:
https://github.com/exercism/docs/blob/master/you-can-help/implement-an-exercise-from-specification.md#configuring-the-exercise

It is a new formatter to format the config files consistently. 
Because the line movings in the files are quite numerous, I'd suggest to merge this now instead of when the next new exercise does it anyway.

### DIFF
--- a/config.json
+++ b/config.json
@@ -1,1048 +1,1046 @@
 {
-  "language": "JavaScript",
   "active": true,
-  "test_pattern": ".*[.]spec[.]js$",
   "exercises": [
     {
-      "uuid": "4756cfc9-7509-4783-8be7-60e3376b8256",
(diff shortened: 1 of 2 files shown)


matthewmorgan:
@vankog so these changes are entirely automated?

Vankog:
Indeed. And it is encouraged by the mentioned doc to run after adding your new exercise to the config. So these changes *will* finally occur in one of the next newly created exercises by someone.
So I think it might be better to commit it separately now instead of mixing it with an unrelated PR.

matthewmorgan:
Thanks @Vankog !!!

PR #406, opened 2017-09-06
reworked pangram spec

Vankog:
The spec file had some strange tests that are not really worth teaching to new devs.
I think we should really have an eye on proper test conventions when reviewing.
(this also is the case for other exercises, but I've only done this for now)

What do you think about this version?
Better? worse? Something I could change?
I think my proposal has a nicer test progression from a TDD point of view and removes unnecessarily duplicated cases and utilizes jasmine's BDD-oriented test description syntax (I hope).

### DIFF
--- a/exercises/pangram/example.js
+++ b/exercises/pangram/example.js
@@ -4,8 +4,9 @@ var notAlpha = /[^a-z]+/gi,
   unique;
 
 var Pangram = function(candidate) {
+  var msg = candidate || '';
   unique = {};
-  cleaned = (candidate.replace(notAlpha, '')).toLowerCase();
+  cleaned = (msg.replace(notAlpha, '')).toLowerCase();
   cleaned.split('').forEach(function (el) {
     unique[el] = true;
   });
--- a/exercises/pangram/pangram.spec.js
+++ b/exercises/pangram/pangram.spec.js
@@ -1,50 +1,61 @@
 var Pangram = require('./pangram');
 
-describe('Pangram()', function()  {
+describe('Tests the Pangram checker.', function()  {
 
-  it('empty sentence', function() {
+  it('can handle an empty sentence', function() {
     var pangram = new Pangram('');
     expect(pangram.isPangram()).toBe(false);
   });
 
-  xit('pangram with only lower case', function()  {
-    var pangram = new Pangram("the quick brown fox jumps over the lazy dog");
-    expect(pangram.isPangram()).toBe(true);
-  });
-
-  xit("missing character 'x'", function()  {
-    var pangram = new Pangram("a quick movement of the enemy will jeopardize five gunboats");
+  xit('handles an undefined message as empty message', function() {
+    var pangram = new Pangram();
     expect(pangram.isPangram()).toBe(false);
   });
 
-  xit("another missing character 'x'", function() {
-    var pangram = new Pangram("the quick brown fish jumps over the lazy dog");
+  xit('recognizes a lower case pangram', function()  {
+    var pangram = new Pangram('thequickbrownfoxjumpsoverthelazydog');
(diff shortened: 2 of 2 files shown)


exercises/pangram/example.js
```diff
@@ -4,8 +4,9 @@ var notAlpha = /[^a-z]+/gi,
   unique;
 
 var Pangram = function(candidate) {
+  var msg = candidate || '';
```
Vankog: had to add a missing null check

exercises/pangram/pangram.spec.js
```diff
@@ -1,50 +1,61 @@
...
+    var pangram = new Pangram('Victor_jagt-zwölf.B0xkämpfer >qu3r< über; den \'großen" Sylter#Deich!');
     expect(pangram.isPangram()).toBe(true);
+    // missing 'd'
+    var pangram = new Pangram('Victor_jagt-zwölf.B0xkämpfer >qu3r< über; einen \'großen" Sylter#Teich!');
+    expect(pangram.isPangram()).toBe(false);
   });
 
-  xit('pangram with non-ascii characters', function()  {
-    var pangram = new Pangram("Victor jagt zwölf Boxkämpfer quer über den großen Sylter Deich.");
+  xit('ignores non-ANSI characters', function()  {
```
Vankog: Trying to bring Unicode to devs attention.

matthewmorgan:
@Vankog thanks for taking this on.  In general my feeling on these changes is that the tests should be brought in to line with the latest [canonical data](https://github.com/exercism/problem-specifications/blob/master/exercises/pangram/canonical-data.json).  Usually the only time we want to vary from that is if there is some JavaScript-specific issue or behavior that ought to be called out.  A good example would be `null` vs `undefined` vs an empty string.  I haven't commented on specific changes, because it looks like this specfile diverges from the canonical one, both in its current form and in your proposed changes.

So, I think the best thing would be:
1. The tests should be updated to match the canonical spec, with the caveat that it's OK and in fact desirable to highlight quirks of this language.
2. If you feel that a test should be added, changed, or removed from the spec, you should make that case in an issue on https://github.com/exercism/problem-specifications.  Once there's consensus there, a PR could be made with the changes.  At that point it would make sense to bring those changes to _specific_ tracks.

Does that make sense to you?

Vankog:
OK, I didn't have the feeling that the canonical tests were taken quite seriously in the tracks. Especially, because the canonicals left some things open to desire. So I took them for inspiration and made sure that they are indeed covered by my new spec. But I also expanded the cases to important edge cases.
So you mean that unless it's something language specific we should really stick only to the canonical cases, without further extension? That feels strange to me, but I see your reasoning to change the canonical definitions first to keep it in sync.

matthewmorgan:
@Vankog actually, what I'm saying is: if you feel strongly about the need to add, remove, or change tests, they should first be added, removed, or changed in the canonical specfile, then in each language track.

I'm not married to the current set of tests, which don't match the canonical specfile, so I would be in favor of changing them, but the right way to do it is by first getting some community consensus on what the right set of tests is.

Does that make sense?

In general I feel the _right_ set of tests not only covers all the cases, but helps the developer recognize the desired input/output patterns.  A map, without dictating internal implementation details, to a working solution.

I think you've identified some weaknesses in the track that also exist in the canonical specfile.

Vankog:
OK, as far as I'm concerned I understood you correctly.
I'll file an issue in the canonical track for discussion.

Vankog:
OK, here is the discussion in the canonical track:
https://github.com/exercism/problem-specifications/pull/893

Vankog:
*i'll open it again when the time is ready (just came back from a vacation anyway).

PR #410, opened 2017-09-13
ported 12 days exercise, resolves issue #391

AdhirRamjiawan:
@rchavarria , @matthewmorgan  please let me know if all is well in this PR, thanks.

### DIFF
--- a/config.json
+++ b/config.json
@@ -1061,7 +1061,7 @@
       "topics": [
         "Control-flow (conditionals)",
         "Control-flow (loops)",
-        "Polymorfism",
+        "Polymorphism",
         "Strings",
         "Pattern recognition",
         "Regular expressions"
--- a/config.json
+++ b/config.json
@@ -1051,6 +1051,21 @@
       "deprecated": true,
       "slug": "point-mutations",
       "uuid": "e9a6b2ea-a67d-4b75-800d-7b46240094ec"
+    },
+    {
+      "uuid": "09e10522-9853-11e7-abc4-cec278b6b50a",
+      "slug": "twelve-days",
+      "core": false,
+      "unlocked_by": "bob",
+      "difficulty": 4,
+      "topics": [
+        "Control-flow (conditionals)",
+        "Control-flow (loops)",
+        "Polymorfism",
+        "Strings",
+        "Pattern recognition",
+        "Regular expressions"
+      ]
     }
   ],
   "foregone": [],
--- a/exercises/twelve-days/example.js
+++ b/exercises/twelve-days/example.js
@@ -19,7 +19,7 @@ var TwelveDays = function() {
   })();
 
   instance.startFromZero = function(oneIndexArray) {
-    var newArray = oneIndexArray.map(item => item - 1);
+    var newArray = oneIndexArray.map(function(item) { return item - 1; });
     return newArray;
   };
 
@@ -29,7 +29,7 @@ var TwelveDays = function() {
   };
 
   instance.multiVerse = function(startIndex, endIndex) {
-    var filteredList = this.verseList.filter((verse, index) => {
+    var filteredList = this.verseList.filter(function (verse, index) {
       if (index >= startIndex && index <= endIndex) {
         return verse;
       }
@@ -37,7 +37,7 @@ var TwelveDays = function() {
     return filteredList.join('\n\n').concat('\n');
   };
 
-  instance.verse = function(...args) {
+  instance.verse = function(args) {
     var indexArray = this.startFromZero(args);
 
     if (args.length === 2) {
--- a/exercises/twelve-days/example.js
+++ b/exercises/twelve-days/example.js
@@ -1,8 +1,5 @@
 var TwelveDays = function() {
-  var instance = this;
-
-  (function() {
-    instance.verseList = [
+  this.verseList = [
       'On the first day of Christmas my true love gave to me, a Partridge in a Pear Tree.',
       'On the second day of Christmas my true love gave to me, two Turtle Doves, and a Partridge in a Pear Tree.',
       'On the third day of Christmas my true love gave to me, three French Hens, two Turtle Doves, and a Partridge in a Pear Tree.',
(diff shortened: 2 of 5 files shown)


exercises/twelve-days/example.js
```diff
@@ -0,0 +1,55 @@
...
+      'On the tenth day of Christmas my true love gave to me, ten Lords-a-Leaping, nine Ladies Dancing, eight Maids-a-Milking, seven Swans-a-Swimming, six Geese-a-Laying, five Gold Rings, four Calling Birds, three French Hens, two Turtle Doves, and a Partridge in a Pear Tree.',
+      'On the eleventh day of Christmas my true love gave to me, eleven Pipers Piping, ten Lords-a-Leaping, nine Ladies Dancing, eight Maids-a-Milking, seven Swans-a-Swimming, six Geese-a-Laying, five Gold Rings, four Calling Birds, three French Hens, two Turtle Doves, and a Partridge in a Pear Tree.',
+      'On the twelfth day of Christmas my true love gave to me, twelve Drummers Drumming, eleven Pipers Piping, ten Lords-a-Leaping, nine Ladies Dancing, eight Maids-a-Milking, seven Swans-a-Swimming, six Geese-a-Laying, five Gold Rings, four Calling Birds, three French Hens, two Turtle Doves, and a Partridge in a Pear Tree.',
+    ];
+  })();
+
+  instance.startFromZero = function(oneIndexArray) {
```
matthewmorgan: Is there a reason you decided to declare these methods as properties of the instance rather than on the TwelveDays prototype? EG:

```javascript
function TwelveDays(){};

TwelveDays.prototype.verseList = [...]
TwelveDays.prototype.startFromZero = function(){...}
```

Either will work in this case, but they're not equivalent.  Just wondering what your thinking was?

exercises/twelve-days/twelve-days.spec.js
```diff
@@ -0,0 +1,106 @@
+var TwelveDays = require('./twelve-days');
+
+describe('TwelveDays', function () {
+  var twelveDaysObject;
+
+  beforeEach(function () {
```
matthewmorgan: I don't believe this is necessary.  You can just instantiate a single instance and use it everywhere.  In fact, I would suggest that you _should_ do so to discourage submissions that might mutate their internal state.

matthewmorgan:
@AdhirRamjiawan looking good!  I have a question and a change request.  Please let me know if I can help further.

exercises/twelve-days/example.js
rchavarria: I didn't see this pattern to create constructor functions. @AdhirRamjiawan, do you have any reference about it? It'd be interesting to read a bit about it.

I don't know about its pros and cons, but I would feel more confortable implementing it as @matthewmorgan suggests

exercises/twelve-days/example.js
```diff
@@ -0,0 +1,55 @@
...
+    return newArray;
+  };
+
+  instance.singleVerse = function(verseIndex) {
+    var verse = this.verseList[verseIndex].concat('\n');
+    return verse;
+  };
+
+  instance.multiVerse = function(startIndex, endIndex) {
+    var filteredList = this.verseList.filter(function (verse, index) {
```
rchavarria: This is not a change request, just to know your opinion. What about chaining a couple of `filter` methods?

```javascript
return this.verseList
  .filter(function(verse, index) { return index >= startIndex })
  .filter(function(verse, index) { return index <= endIndex })
  .join('\n\n')
  .concat('\n')
```

exercises/twelve-days/example.js
AdhirRamjiawan: @matthewmorgan , nope, in this case no particular reason why I've created them as instance methods. I've just been used to creating functions in this way. I will change it to use the prototype.

@rchavarria I'll try to look for some sites to ref this however it's just been something I've been doing for a while now. When working with some front end frameworks like Angular/Jquery I occasionally found myself needing to preserve "this" in the scope it's in especially when nesting functions. I just use it as a way to get functionality into a JS object without having any unintended side effects with using the prototype chain.

exercises/twelve-days/twelve-days.spec.js
AdhirRamjiawan: Okay, so declaring twelveDaysObject as a global should be fine?

exercises/twelve-days/example.js
AdhirRamjiawan: Your suggestion reads much better. I think the filter method just needs the callback to return a "truthy" value or not to decide whether to keep the element. Returning the boolean result instead of  `return verse;` makes it easier to understand.

AdhirRamjiawan:
@matthewmorgan @rchavarria I've commit a few changes as discussed above

exercises/twelve-days/twelve-days.spec.js
matthewmorgan: You're never mutating anything in the object, so it would be fine.

config.json
```diff
@@ -1051,6 +1051,21 @@
...
+    {
+      "uuid": "09e10522-9853-11e7-abc4-cec278b6b50a",
+      "slug": "twelve-days",
+      "core": false,
+      "unlocked_by": "bob",
+      "difficulty": 4,
+      "topics": [
+        "Control-flow (conditionals)",
+        "Control-flow (loops)",
+        "Polymorfism",
```
matthewmorgan: Should read "Polymorphism".

exercises/twelve-days/example.js
```diff
@@ -0,0 +1,51 @@
...
+      'On the eleventh day of Christmas my true love gave to me, eleven Pipers Piping, ten Lords-a-Leaping, nine Ladies Dancing, eight Maids-a-Milking, seven Swans-a-Swimming, six Geese-a-Laying, five Gold Rings, four Calling Birds, three French Hens, two Turtle Doves, and a Partridge in a Pear Tree.',
+      'On the twelfth day of Christmas my true love gave to me, twelve Drummers Drumming, eleven Pipers Piping, ten Lords-a-Leaping, nine Ladies Dancing, eight Maids-a-Milking, seven Swans-a-Swimming, six Geese-a-Laying, five Gold Rings, four Calling Birds, three French Hens, two Turtle Doves, and a Partridge in a Pear Tree.',
+    ];
+
+  
+};
+
+TwelveDays.prototype.startFromZero = function(oneIndexArray) {
+  var newArray = oneIndexArray.map(function(item) { return item - 1; });
+  return newArray;
```
matthewmorgan: This is a minor not, but you don't need to assign the mapped value to `newArray`, you could just return it:

```javascript
TwelveDays.prototype.startFromZero = function(oneIndexArray) {
    return oneIndexArray.map(function(item) { return item - 1; });
}
```

matthewmorgan:
@AdhirRamjiawan I found one more thing that should be changed (a spelling error.) So, to recap:
1. Create one (and only one) `TwelveDays` object in the test suite.
2. Fix that spelling error!

Other than that, feel free to act on my other feedback or not, as you wish. 

Otherwise this is looking good to me!  @rchavarria ?

exercises/twelve-days/twelve-days.spec.js
```diff
@@ -0,0 +1,105 @@
+var TwelveDays = require('./twelve-days');
+
+describe('TwelveDays', function () {
+  var twelveDaysObject = new TwelveDays();
+
+  beforeEach(function () {
```
rchavarria: If it's not needed, I would just delete it

rchavarria:
@AdhirRamjiawan many many thanks for working on this

exercises/twelve-days/twelve-days.spec.js
AdhirRamjiawan: It doesn't seem as though it's possible to remove the new TwelveDays object on line 4. Even though line one calls the require function, a new instance would need to be made for the prototype members to be available. Any suggestions?

AdhirRamjiawan:
@rchavarria , @matthewmorgan I've only been able to fix the spelling error on the config.json file. Having some issues with removing the TwelveDays object on line 4

exercises/twelve-days/twelve-days.spec.js
matthewmorgan: @AdhirRamjiawan what we meant is that the `beforeEach` method is no longer needed.  You are correct that the `twelveDaysObject` must still be declared.

Sorry if that was unclear :)

matthewmorgan:
@AdhirRamjiawan thanks for making that last change.  Do you know why CI is failing?  I'm not able to see the error report.

AdhirRamjiawan:
@matthewmorgan , not sure why it failed. It's pretty weird. I've pushed a dummy commit and seems to pass now.

PR #411, opened 2017-09-13
ESLint with Airbnb Style Guide

rootulp:
We can now lint exercises with `yarn lint`
Initial run returns: `✖ 17318 problems (15822 errors, 1496 warnings)`


rootulp:
@rchavarria Would you mind reviewing?

rootulp:
Or @matthewmorgan would you mind reviewing?

rchavarria:
Whoah! 17k errors! And lots of changes. This will be a hard PR. 

I expected lots of errors, because many people has contributed to the repo, so there are lots of styles in the source code.

But, there might be another reason. I saw that the `--fix` flag has changed `var` with `const` or `function` with fat arrow functions `() => {}`. The airbnb style is right for the ecmascript track, but I wonder if it's right for this track as it is now. For now, the track is focused on prior-ES2015 features (no `const`, no arrow functions,...). Maybe, there is an option in the airbnb style to just check for ES5 features.

And that, raises THE question again: should this track move to ES2015 features? Recently, Node.js v8.5 anounced that it supports ES modules. So, the time to move to ES2015 is coming. But not right now, or at least, not in this PR.

@rootulp, could you check if there is an option to lint just ES5 features?

BTW, please, don't spend too much time fixing errors manually, that'll be too much work for a single person. I think this linting thing is raising very interesting concepts to talk and discuss, that should be addressed before fixing code style errors.

Thanks for this great effort!

rootulp:
- Yes, we can probably use [eslint-config-airbnb-es5](https://github.com/1hella/eslint-config-airbnb-es5) or something similar
- Agreed on not fixing errors manually
- I don't currently have strong opinions on whether this track should adopt ES2015 features. Seems like we _should_ but then I don't see the purpose of having a Javascript track and an ECMAscript track. @rchavarria can you make a case for one or the other?

matthewmorgan:
@rootulp At the moment we're planning on keeping this track supporting ES5 features, and the EcmaScript track ES6 and beyond.  

The  `eslint-config-airbnb-es5` approach looks good to me.  @rchavarria ?

rootulp:
After switching to es5 lint rules and running eslint with `--fix` we got down to:
`✖ 4789 problems (3256 errors, 1533 warnings)`
Then added the jasmine environment and we're now at:
`✖ 2784 problems (1251 errors, 1533 warnings)`

Initial impression of lint output:
- A lot of errors come from big-integer.js 
- A lot of warnings come from [func-names ](https://eslint.org/docs/rules/func-names)

rchavarria:
It seems `eslint-config-airbnb-es5` isn't official. But at least it's a packaged set of rules. And it outputs much less errors. I'm ok to go with it.

About the `func-names` rule, I'd suggest to configure it to suppress it, so that we don't get errors/warnings due to that rule.

About using `npm` or `yarn`, I'm ok if people wants to use `yarn`, but I think we should favor the most common option, `npm`. I hope both tools will be compatible for a long time.

rootulp:
After disabling `func-names`:
`✖ 1292 problems (1251 errors, 41 warnings)`

I believe `npm` and `yarn` are fairly interchangeable but happy to standardize on `npm` if that's what the track prefers.

rchavarria:
I feel good about accepting this PR about linting.

@rootulp - why did you remove `$` in the docs? I remember discussing it recently, but I don't remember it was in this track or in ecmascript track, sorry. I would prefer to have it in a separate PR but it's not a big deal for me, so don't worry, don't revert the commit yet.

@exercism/ecmascripts, is it ok to accept this?

joelwallis:
Questions:

- is there a good reason to change the Node.js version on `.travis.yml`? It doesn't seem to be in the scope of this PR by first sight.
- @rootulp seem to be using Yarn as package manager. Although I'm starting to use it on my personal projects, should we keep it as the package manager for Exercism JS-related tracks (ECMAScript and JavaScript)? I like Yarn but don't think we need to add an additional installation step to our users (we'll need to add it to the [installation instructions for the ECMAScript track](http://exercism.io/languages/ecmascript/installation).

Besides that, it looks very good to me.

joelwallis:
Another question: is Travis now running ESLint in the codebase as an additional CI step?

rootulp:
@rchavarria I removed the `$` to be consistent with other language tracks (see https://github.com/exercism/ecmascript/pull/352)

@joelwallis I upgraded to the latest stable release of Node in `.travis.yml` b/c a build [failed](https://travis-ci.org/exercism/javascript/builds/275253166?utm_source=github_status&utm_medium=notification) due to lint --fix applying ECMAscript 6 features (use of `let`). We _could_ downgrade back to the earlier version of Node but I see no harm in being on the latest stable release.

@joelwallis I use yarn but there are both `package-lock.json` & `yarn.lock` files. Running `npm install` will still work. I'm happy to remove the `yarn.lock` file if this causes confusion.

rootulp:
Not yet.

joelwallis:
Cool. Let's keep the latest version of Node.js. We definitely need to be able to `let` a variable, so 😄

About Yarn, I would prefer to avoid it, in favor of npm (as well as keeping the `package-lock.json`). Adding an additional installation step doesn't look like a good choice.

README.md
```diff
@@ -6,20 +6,23 @@ Exercism exercises in JavaScript
...
 
-### Single Assignment
+### Unit Tests: Single Assignment
 
-    % make test-assignment ASSIGNMENT=wordy
+    make test-assignment ASSIGNMENT=wordy
+
+### Code Style
+
+    yarn lint
```
rchavarria: I think we agree on favoring `npm` over `yarn`. But it'll be interesting to allow contributors to use `yarn`. This doc could be like:

> Lint the code with `npm run lint`. If you use `yarn`, the command is `yarn lint`.

joelwallis: I believe Yarns users are aware that it's possible to run npm scripts with the `yarn` executable. Someone who's not using the official Node.js package manager must have solid reasons [and comprehension of them] to be using it. Besides that, it's impossible for a contributor to have `yarn` but not `npm` available, since npm is bundled in Node.js.

Do you really think that a mention to Yarn in docs is something that worth keeping, @rchavarria?

rchavarria: I don't have a strong feeling about it. I think we all agree that `npm` should be the tool to be exposed in the docs. `yarn` seems to be quite spread. I don't mind if the docs make some reference to it, like:

>  If you prefer, you can use `yarn` :wink: :wink:

rootulp:
Removed use of yarn. Ready to be merged? @rchavarria @joelwallis

rchavarria:
Yep, I'd ask you to squash all the commits into a single one, but I think this PR contains a very big effort. Thanks a lot @rootulp for your great work.

There are only thousands of lint errors to fix :rofl: 

I'm going to merge this.

joelwallis:
Hey @rootulp, thank you very much for this awesome contribution! 🎉

PR #415, opened 2017-10-09
protein-translation: add to track

Smarticles101:


Smarticles101:
This should be ready for review now. Doesn't look like the javascript track is super active, so review/merge when you can :) I'll try to contribute some more, especially since I need to brush up on my javascript.

exercises/protein-translation/example.js
```diff
@@ -0,0 +1,66 @@
...
+function translate(rnaStrand) {
+  var proteins = [];
+
+  if (rnaStrand) {
+    for (var i = 0; i < rnaStrand.length; i+=3) {
+      var protein = getProtein(rnaStrand.substring(i, i + 3));
+
+      if (protein) {
+
+        if (protein === "STOP")
```
matthewmorgan: We're using the AirBnb JS style guide on this track, so your `if` statements should have [braces please](https://github.com/airbnb/javascript#blocks--braces)!

Smarticles101: Fixed!

matthewmorgan: Thanks @Smarticles101 !

PR #416, opened 2017-10-09
Added Transpose Exercise

kk-r:
@matthewmorgan Transpose Exercise completed. Please review


config.json
```diff
@@ -1080,6 +1080,21 @@
         "Pattern recognition",
         "Regular expressions"
       ]
+    },
+    {
+      "core": false,
```
matthewmorgan: I believe this exercise must either be marked as `"core": true` or be `unlocked` by another exercise.  May I suggest the `matrix` exercise as the one to unlock this one?

exercises/transpose/example.js
```diff
@@ -0,0 +1,15 @@
...
+    line.split('').map(function (value, key) {
+      if (typeof result[key] === 'undefined') {
+        result[key] = new Array(lineNo + 1).join( ' ' );
+      }
+      result[key] += value;
+    });
+    return result;
+  }, []).join('\n');
+}
+
```
matthewmorgan: Can you please run ESLint against your PR?  There are a few small style issues with the code, and I don't want to waste your time calling them out individually.  But if you use the linter it can fix your code automatically!

You can lint a _single_ exercise dir like this:
1. Make sure you're in the root level of the javascript repo
2. Make sure you've run `npm install`
3. run `./node_modules/.bin/eslint exercises/transpose/ --fix`

This will fix any small style issues (EG whitespace, braces, etc) that it can, and display errors for anything it cannot fix.  Big time-saver.

exercises/transpose/transpose.spec.js
```diff
@@ -0,0 +1,65 @@
...
+  });
+
+  xit('test two characters in a row', function () {
+    var input = 'A1';
+    var expected = 'A\n1';
+    expect(transpose(input)).toEqual(expected);
+  });
+
+  xit('test two characters in a column', function () {
+    var input = 'A\n1';
```
matthewmorgan: When I look at the `problem-specifications` repo, it may be slightly unclear what the expected form of the result is.  However, the examples all take arrays as input and expect arrays as results.

I think it would be best if you altered your solution to match the API described in the problem-spec repo, rather than have one that consumes and returns strings.

kk-r:
@matthewmorgan Done. Please review and let me know :+1:

config.json
rchavarria: I don't remember how many exercises are marked as `core`, so I'd prefer having it as non-core. `unlocked` by `matrix` seems ok to me.

exercises/transpose/transpose.spec.js
rchavarria: Yep, I agree with @matthewmorgan, `transpose` should accept an `array` as parameter and return an `array` as a result.

@kk-r, you'll find the canonical data @matthewmorgan is talking about here: https://github.com/exercism/problem-specifications/blob/master/exercises/transpose/canonical-data.json

config.json
kk-r: @matthewmorgan @rchavarria  I updated the config file -  `"core": false, "unlocked_by": "matrix"`.

rchavarria:
I'm going to merge this. I'm missing the README file too, but I'm ready to generate it (I think).

PR #417, opened 2017-10-12
reverse string exercise

eliaahadi:
I added new reverse string exercise, I'm happy to edit it.

PR #418, opened 2017-10-12
Update config.json

eliaahadi:
file: added reverse-string config info


matthewmorgan:
@eliaahadi this would need to be part of the exercise PR itself.  It is required for the CI suite to run and test the code before merging.

Let's figure out what is the next step for #417 and then we can make this change there.

PR #425, opened 2017-10-14
Fixed link to CLI download page.

cruxicheiros:
It was http://exercism.io/cli, which resulted in a 404.


matthewmorgan:
nice catch!

Issue #399, opened 2017-08-23
Fix code styles of all example solutions


joelwallis:
~~**Blocked by #373**~~

---

We already have ~~an issue to add~~ added linting capabilities to our codebase using eslint. Now we need to make all exercise files compliant with the linting rules. 

We will be spitting it up into exercises and would like to have a separate PR for each exercise.

**Ready to go:**
Below is a list of all files that need linting along with the ones which are already linted (checked).
See the full list of exercises here: https://github.com/exercism/javascript/tree/master/exercises

- [x] accumulate
- [x] acronym
- [x] all-your-base
- [x] allergies
- [x] alphametics
- [x] anagram
- [x] atbash-cipher
- [x] beer-song
- [x] binary
- [x] binary-search
- [x] binary-search-tree
- [x] bob
- [x] bowling
- [x] bracket-push
- [x] change
- [x] circular-buffer
- [x] clock
- [x] collatz-conjecture
- [x] connect
- [x] crypto-square
- [x] custom-set
- [x] diamond
- [x] difference-of-squares
- [x] diffie-hellman
- [x] etl
- [x] flatten-array
- [x] food-chain
- [x] forth
- [x] gigasecond
- [x] grade-school
- [x] grains
- [x] hamming
- [x] hello-world
- [x] hexadecimal
- [x] isbn-verifier
- [x] isogram
- [x] kindergarten-garden
- [x] largest-series-product
- [x] leap
- [x] linked-list
- [x] list-ops
- [x] luhn
- [x] matrix
- [x] meetup
- [x] minesweeper
- [x] nth-prime
- [x] nucleotide-count
- [x] ocr-numbers
- [x] octal
- [x] palindrome-products
- [x] pangram
- [x] pascals-triangle
- [x] perfect-numbers
- [x] phone-number
- [x] pig-latin
- [x] point-mutations
- [x] prime-factors
- [x] protein-translation
- [x] proverb
- [x] pythagorean-triplet
- [x] queen-attack
- [x] raindrops
- [x] reverse-string
- [x] rna-transcription
- [x] robot-name
- [ ] robot-simulator
- [x] roman-numerals
- [x] run-length-encoding
- [x] saddle-points
- [x] say
- [x] scrabble-score
- [x] secret-handshake
- [x] series
- [x] sieve
- [ ] simple-cipher
- [x] simple-linked-list
- [x] space-age
- [x] strain
- [x] sublist
- [x] sum-of-multiples
- [x] transpose
[...]

PR #427, opened 2017-10-15
Fix Meetup code

Radix16:
Correct the 'meetup' exercise for formatting and syntax error as indicated by the linter. #399

### DIFF
--- a/exercises/meetup/example.js
+++ b/exercises/meetup/example.js
@@ -1,43 +1,40 @@
+'use strict';
+
 function MeetupDayException(message) {
   this.message = message;
   this.name = 'MeetupDayException';
 }
 
-function meetupDay(year, month, day_of_week, which) {
-  'use strict';
+function meetupDay(year, month, dayOfWeek, which) {
+  var candidates = _getCandidates(year, month, dayOfWeek);
+  var res;
+  var day;
 
-  var candidates = _getCandidates(year, month, day_of_week),
-    d,
-    i,
-    res;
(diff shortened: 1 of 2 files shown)


exercises/meetup/example.js
```diff
@@ -56,15 +53,17 @@ function _getDayIndex(day) {
     'saturday': 6
   };
 
-  day = day.toLowerCase();
+  var temp = day.toLowerCase();
```
matthewmorgan: Hmm.  Can we find a more descriptive variable name than `temp` for this?  Otherwise, I'd rather see:

```javascript
return daysInd[day.toLowerCase()];
```

exercises/meetup/example.js
```diff
@@ -56,15 +53,17 @@ function _getDayIndex(day) {
...
 
-  return daysInd[day];
+  return daysInd[temp];
 }
 
 function _find(ary, callback) {
   for (var i = 0; i < ary.length; i++) {
     if (callback(ary[i], i, ary)) { return ary[i]; }
   }
+  return null;
```
matthewmorgan: `return undefined` is implicit, and preferred by the [style guide](https://github.com/airbnb/javascript#arrows--implicit-return), so let's not do this.

matthewmorgan:
@Radix16 Thanks!

BTW I think I may have misinterpreted the implicit return issue on AirBnB's guide.  It seems they're neutral on it, not recommending it.  But thank you for making these changes.

PR #431, opened 2017-10-16
Shows joelwallis as maintainer in Nextercism

joelwallis:


rchavarria:
No link? No bio? 😢 

I think at least a short bio would help the website to have a nice structure. Just a phrase telling who you are or why you like programming or just anything about you.

joelwallis:
I'll add a bio. There's no need to add a link, my GitHub profile will be shown as default URL.

joelwallis:
Done!

Name, URL, etc., will fallback to my info at GitHub, so `bio` is the only key that seems necessary.

rchavarria:
Great!!

About the URL is just a matter of taste. But I think that the bio should contain something about the person. That's why I *complained* 😉

joelwallis:
Thanks, Rubén!

PR #433, opened 2017-10-17
fix the change exercise

Radix16:
fix the change exercise.  #399 
Only the example.js file required changes.  The change.spec.js does not have any errors reported by the linter.


rchavarria:
Many thanks for taking care of this :tada:

PR #434, opened 2017-10-18
fix the protein-translation exercise

Radix16:
Correct the example.js and protein-translation.spec.js files as indicated by the linter.  #399


rchavarria:
130 errors less to worry about!

image

You rock @Radix16!! Really appreciated.

PR #435, opened 2017-10-20
fix the twelve-days exercise

Radix16:
Fix the twelve-days exercise as indicated by the linter.  No changes were needed in the twelve-days.spec.js file.  #399


matthewmorgan:
Thanks @Radix16!  Cranking 'em out...

PR #437, opened 2017-10-22
fix the grains exercise

Radix16:
I added .eslignore files to the grains exercise directory and project root.   Also deleted the big-integer-spec.js test file as it doesn't make sense to keep a test for a vendor included library.  A user won't be able to get to it anyway without download the project source.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -14,7 +14,8 @@ exercises/diamond
 exercises/flatten-array
 exercises/food-chain
 exercises/grade-school
-exercises/grains
+exercises/grains/big-integer.js
+exercises/grains/big-integer.spec.js
 exercises/kindergarten-garden
 exercises/leap
 exercises/linked-list
--- a/exercises/grains/example.js
+++ b/exercises/grains/example.js
@@ -4,7 +4,7 @@ var BigInt = require('./big-integer');
  * @author github.com/nonsensery
  * @class Grains
  *
- * Computes the number of grains on the squares of a
+ * @classdesc Computes the number of grains on the squares of a
  * chess board, starting with one grain on the first
  * square, and doubling with each successive square.
  */
@@ -14,16 +14,19 @@ function Grains() {
 
 /**
  * Gets the number of grains on the nth square.
+ * @param {number} num - the value to square
+ * @return {number} the square of num
  */
 Grains.prototype.square = function (num) {
-  return BigInt(2).pow(num - 1).toString();
+  return new BigInt(2).pow(num - 1).toString();
 };
 
 /**
  * Gets the total number of grains on all squares.
+ * @return {string} the total
  */
 Grains.prototype.total = function () {
-  var total = BigInt(0);
+  var total = new BigInt(0);
 
   for (var squareNum = 1; squareNum <= 64; ++squareNum) {
     total = total.add(this.square(squareNum));


Radix16:
#399

exercises/grains/big-integer.spec.js
```diff
@@ -1,59 +0,0 @@
-var BigInt = require('./big-integer');
```
rchavarria: I don't think this file should be deleted. It serves as documentation for the `big-integer` library. If it contains ESLint errors, it could be ignored like `big-integer.js`

.eslintignore
```diff
@@ -0,0 +1,4 @@
+#eslint will ignore these files
```
rchavarria: `/.eslintignore` or `/exercises/grains/.eslintignore` is probably redundant. I would favor the one on `/exercises/grains/`, because it's more specific.

Have you tried with only one of these files?

Radix16:
@matthewmorgan The linter only uses the .eslintignore file from the current directory where it is executed.   Thus, we should keep the .eslintignore file at the root so that the .eslintignore file is available when the linter is executed for the whole project.  The other one is a convenience for those who run the linter for just the grains exercise.   I created them in both locations to cover both usages.   However I will remove the .eslintignore file in the grains exercise, as you suggest.   Please see https://eslint.org/docs/user-guide/configuring#using-eslintignore-in-packagejson.   BTW, I also tried the package.json file method of ignoring the files which didn't work any better.  If you want a different implementation please let me know.  

Also, I will bring back the spec file.

rchavarria:
In the https://github.com/exercism/ecmascript track we decided to put all ESLint config in `package.json`. Here, the config is outside it. I would follow the current approach here, so I would keep the `.eslintignore` outside the `package.json` file. That would make this PR more focused.

I'll open a issue to discuss if we should put ESLint config inside `package.json`

Radix16:
@rchavarria the grains.spec.js file is back and I kept only one of the .eslintignore files.  I keep the one at the project root because I the linter uses only looks in the current directory the .eslintignore file.

exercises/grains/example.js
```diff
@@ -1,29 +1,33 @@
-var BigInt = require('./big-integer');
+var bigInt = require('./big-integer');
```
rchavarria: Hmm. I really like `BitIng`, starting with uppercase.

I undersand your change, ESLint is complaining about that. Have you tried to use `new BigInt(2)` instead of `BigInt(2)`? It might be the same for this exercise and for the use of `BigInt` in this exercise. But, I think `BigInt` is a constructor function, so I'd prefer using `new` instead of starting it in lowercase.

tejasbubane:
@Radix16 We have now added eslint check in the CI and an [`eslintignore` file](https://github.com/exercism/javascript/blob/master/.eslintignore) with list of unimplemented exercises. We now expect contributors fixing the lint errors, to remove that particular exercise (`diamond` and `sum-of-multiples` in your case). Would you mind [rebasing your branch](https://www.atlassian.com/git/tutorials/rewriting-history/git-rebase) on master and removing them from the eslintignore?

Please feel free to let us know if this sounds unfamiliar you are stuck with some problem.

matthewmorgan:
@Radix16 this is looking good, but would you mind addressing @rchavarria 's question about `bigInt` vs `BigInt`?  

Seems like if we can get that question satisfied, we are ready to merge.  Thanks!

Radix16:
@matthewmorgan Eeek!  Yes I will make the change back to Bigint.   I thought I had, but I may have lost that version somewhere.   I will fix it.   Thanks for finding the omission.

Radix16:
@matthewmorgan I fixed the BigInt

matthewmorgan:
Thanks!  I wasn't sure if you and Rubén were on the same page, but I just didn't want it to fall through the cracks. 

We appreciate all your work!

PR #439, opened 2017-10-23
Implement exercise: zipper

duailibe:


exercises/zipper/example.js
```diff
@@ -0,0 +1,104 @@
...
+        value: this.tree.value,
+        left: this.tree.left,
+        right: right
+      },
+      this.trail
+    );
+  };
+
+  return Zipper;
+})();
```
matthewmorgan: The IFFE is not needed here because the module system controls the scope.

You can just define `Zipper` like this:
```javascript
var Zipper = function (tree, trail){
    this.tree = tree;
    this.trail = trail;
};

Zipper.prototype.fromTree = function(tree){
...
};
```

and so on.  Having the IFFE with the constructor inside the IFFE is a bit confusing and since we don't need it at all, we should simplify the code and remove it.

Functions that shouldn't be exposed as part of the public API (EG `fromTrail()` ) can simply be included in the module, but outside of the `Zipper` definition itself.

exercises/zipper/zipper.spec.js
```diff
@@ -0,0 +1,78 @@
...
+  return bt(value, null, null);
+}
+
+var t1 = bt(1, bt(2, null, leaf(3)), leaf(4));
+var t2 = bt(1, bt(5, null, leaf(3)), leaf(4));
+var t3 = bt(1, bt(2, leaf(5), leaf(3)), leaf(4));
+var t4 = bt(1, leaf(2), leaf(4));
+var t5 = bt(1, bt(2, null, leaf(3)), bt(6, leaf(7), leaf(8)));
+var t6 = bt(1, bt(2, null, leaf(5)), leaf(4));
+
```
matthewmorgan: These variables seem to be the concern of the `describe` block, and (if I understand your intention) would not be used in a different `describe` block, so shouldn't they be declared inside the block?

The functions seem to be a different matter: they seem to be the sort of thing you might use repeatedly across `describe` blocks, so they seem to be in the right place.

exercises/zipper/zipper.spec.js
```diff
@@ -0,0 +1,78 @@
...
+describe('Zipper', function () {
+  var zipper;
+
+  beforeEach(function () {
+    zipper = Zipper.fromTree(t1);
+  });
+
+  it('data is retained', function () {
+    expect(zipper.toTree(t1)).toEqual(t1);
```
matthewmorgan: Nice!  You remembered to unskip the first test...

exercises/zipper/README.md
```diff
@@ -0,0 +1,55 @@
...
+  Execute the tests with:
+
+      jasmine <exercise-name>.spec.js
+
+  Replace `<exercise-name>` with the name of the current exercise. E.g., to
+  test the Hello World exercise:
+
+      jasmine hello-world.spec.js
+
+  In many test suites all but the first test have been skipped.
```
matthewmorgan: Since there's only one suite (IE `describe` block) and you skipped all but the first test, could we change this to read:
> All tests but the first have been skipped.

duailibe:
Thanks for the review!

I think I addressed all your comments, let me know if something else needs changing :)

exercises/zipper/example.js
```diff
@@ -1,104 +1,100 @@
...
+  var last = trail[0];
+  return rebuildTree(fromTrail(tree, last), trail.slice(1));
+}
+
+var Zipper = function (tree, trail) {
+  this.tree = tree;
+  this.trail = trail;
+};
+
+Zipper.fromTree = function (tree) {
```
matthewmorgan: Should this be declared on the prototype, like all the other functions?  Or is this a deliberate choice?

duailibe: This is meant to be a "constructor", like `Array.from()`. That means the following should not work:
```js
zipper.left().right().fromTree(someTree)
```

Does that make sense?

rchavarria: The `README` describes `fromTree` at the same level as `toTree`,... But it makes sense to have it like a kind of a named constructor. So, I like `Zipper.fromTree` and the following doesn't make sense for me:

```javascript
zipper.left().right().fromTree(someTree)
```

because, is it creating a tree from `someTree` or from `zipper.left().right()`?

duailibe: Yes, @rchavarria, exactly what I meant :)

matthewmorgan:
@rchavarria look OK to you?  I feel this is ready to merge.

rchavarria:
Sure, it's great! Thanks @duailibe!

PR #440, opened 2017-10-24
Fix code style in prime-factors

PakkuDon:
See #399 
Only needed to update `example.js`, no errors were reported in `prime-factors.spec.js`


rchavarria:
Very well done!! 9 ESLint error less, awesome!

Many thanks for this great work.

PR #444, opened 2017-10-29
Lint Fixes for Beer Song Example

lougeniaC64:
Fixed beer-song example as request in Issue #399
- spec doc had no lint errors


rchavarria:
Fabulous! Thanks for this contribution. What about picking another exercise of #399 and fix its ESLint errors?

PR #445, opened 2017-10-29
Fixed Lint Errors in Diffie-Hellman

lougeniaC64:
Lint error fixes for diffie-hellman spec and example files as per #399.


exercises/diffie-hellman/example.js
```diff
@@ -19,14 +19,14 @@ var DiffieHellman = function(p, g){
     return Math.pow(theirPublicKey, ourPrivateKey) % this.p;
   };
 
-  this.validateInitialArguments = function (p, g){
```
rchavarria: I really like the names `p` and `g`. I don't know what they mean exactly, but they're used in the `README` so they use the language of the specification.

I know this PR is not changing the method, the method was like this before, but I have an idea that may improve the code. Just let me know what do you think: instead of passing `p` and `g` as parameters to `validateInitialArguments`, could that method use `this.p` and `this.g`? what do you think of that refactor?

matthewmorgan: @rchavarria `p` stands for "prime modulus" and `g` stands for "generator".  I am not sure that helps anyone, but that is conventionally what they mean.

exercises/diffie-hellman/example.js
```diff
@@ -19,14 +19,14 @@ var DiffieHellman = function(p, g){
...
+  this.validateInitialArguments = function (this.p, this.g) {
```
lougeniaC64: @rchavarria @matthewmorgan Let me know if this change is what you had in mind.

lougeniaC64: Nevermind.  Tests are failing with that change.

lougeniaC64: So maybe `prime modulus` instead of this.p and `generator` instead of this.g?  I don't want to do anything convoluted to resolve this issue.

exercises/diffie-hellman/example.js
rchavarria: Yep @matthewmorgan, that helps, thanks.

That was not the change I requested, but at least parameters have a different name (lint error is fixed) and the names are readable. So, I'm ok with this change too. :+1:

lougeniaC64: @rchavarria My apologies.  I tried to make the name changes you suggested, but that broke some of the tests.  I should have looked into why the errors were happening, but I was pressed for time.

rchavarria: I have to apology too. I didn't see your comments about `this.p` change causing tests to fail (GitHub hides comments on *outdated* code).

In the end, the code becomes more readable, and that's nice. Thanks a lot!!

Issue #447, opened 2017-10-30
Change installation link in exercise READMEs


tejasbubane:
Each [exercise](https://github.com/exercism/javascript/tree/master/exercises) has a [README file with a setup link in it](https://github.com/exercism/javascript/blob/master/exercises/accumulate/README.md#setup).

The link points to http://exercism.io/languages/javascript/ but the setup instructions are in http://exercism.io/languages/javascript/installation. 

**Task:** Change the link to the correct installation link in README files of `all` exercises.

Evalle:
Hey, @tejasbubane may I take a look at it?

tejasbubane:
@Evalle Go for it!

PR #449, opened 2017-10-30
ISSUE-447 - Fix the path to setup page for each exercise

Evalle:
Fixes the path to setup page for each exercise, see https://github.com/exercism/javascript/issues/447 for more information.


rchavarria:
Both #448 and #449 implement #447.

This is a pity, but we need to choose one. I would say #448, only because the PR was submitted before. But, I don't have a strong feeling on which one should be accepted.

Anyway, thanks @RandyCupic and @Evalle for you work, your effort it's highly appreciated.

PR #451, opened 2017-10-31
Fix crypto square, largest series product, sieve examples

saramartinez:
Part of https://github.com/exercism/javascript/issues/399

Fix `error  Split 'var' declarations into multiple statements  one-var` errors to clean up three examples. These were the only errors on these three examples.

### DIFF
--- a/exercises/crypto-square/example.js
+++ b/exercises/crypto-square/example.js
@@ -22,7 +22,8 @@ module.exports = function (input) {
 
   this.ciphertext = function () {
     var textSegments = this.plaintextSegments();
-    var i, j;
+    var i;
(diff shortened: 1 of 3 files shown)


rchavarria:
Many thanks for fixing those errors! :tada: 

The idea should be to fix errors for one exercise on each PR. There is no problem here, because there are just a few errors. Next time, don't worry, open as many PRs as exercises are involved, one PR for each exercise.

Then, there will be so many PRs to review, that I and the rest of the maintainers will be overwhelmed with them and we'll ask you to group several exercises in the same PR. So, please, bear with us :pray:

PR #452, opened 2017-10-31
Fix Diamond and Sum of Multiples

saramartinez:
Part of https://github.com/exercism/javascript/issues/399

Fixes lint errors for Diamond and Sum of Multiples examples.

Diamond: 

```
/javascript/exercises/diamond/diamond.spec.js
   7:5   error  'result' is not defined  no-undef
   8:46  error  'result' is not defined  no-undef
  12:5   error  'result' is not defined  no-undef
  17:46  error  'result' is not defined  no-undef
  21:5   error  'result' is not defined  no-undef
  30:46  error  'result' is not defined  no-undef

/javascript/exercises/diamond/example.js
  8:14  error  'i' is already defined  no-redeclare

✖ 7 problems (7 errors, 0 warnings)
```

Sum of Multiples:
```
/javascript/exercises/sum-of-multiples/sum-of-multiples.spec.js
   5:12  error  A function with a name starting with an uppercase letter should only be used as a constructor  new-cap
   9:12  error  A function with a name starting with an uppercase letter should only be used as a constructor  new-cap
  13:12  error  A function with a name starting with an uppercase letter should only be used as a constructor  new-cap
  17:12  error  A function with a name starting with an uppercase letter should only be used as a constructor  new-cap
  21:12  error  A function with a name starting with an uppercase letter should only be used as a constructor  new-cap
  25:12  error  A function with a name starting with an uppercase letter should only be used as a constructor  new-cap
  29:12  error  A function with a name starting with an uppercase letter should only be used as a constructor  new-cap
  33:12  error  A function with a name starting with an uppercase letter should only be used as a constructor  new-cap
  37:12  error  A function with a name starting with an uppercase letter should only be used as a constructor  new-cap

✖ 9 problems (9 errors, 0 warnings)
```

### DIFF
--- a/exercises/diamond/example.js
+++ b/exercises/diamond/example.js
@@ -5,8 +5,8 @@ var Diamond = function () {
     for (var i = 0; i <= inputIndex; i++) {
       output += getLine(inputIndex, i);
     }
-    for (i = inputIndex - 1; i >= 0; i--) {
-      output += getLine(inputIndex, i);
+    for (var j = inputIndex - 1; j >= 0; j--) {
+      output += getLine(inputIndex, j);
     }
     return output;
   };
--- a/exercises/diamond/example.js
+++ b/exercises/diamond/example.js
@@ -5,7 +5,7 @@ var Diamond = function () {
     for (var i = 0; i <= inputIndex; i++) {
       output += getLine(inputIndex, i);
     }
-    for (var i = inputIndex - 1; i >= 0; i--) {
+    for (i = inputIndex - 1; i >= 0; i--) {
       output += getLine(inputIndex, i);
     }
     return output;
--- a/exercises/sum-of-multiples/sum-of-multiples.spec.js
+++ b/exercises/sum-of-multiples/sum-of-multiples.spec.js
@@ -1,39 +1,39 @@
-var sumOfMultiples = require('./sum-of-multiples');
+var SumOfMultiples = require('./sum-of-multiples');
 
-describe('sumOfMultiples', function () {
+describe('SumOfMultiples', function () {
   it('to 1', function () {
-    expect(sumOfMultiples([3, 5]).to(1)).toBe(0);
+    expect(new SumOfMultiples([3, 5]).to(1)).toBe(0);
   });
 
   xit('to 3', function () {
-    expect(sumOfMultiples([3, 5]).to(4)).toBe(3);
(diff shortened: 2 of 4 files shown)


exercises/diamond/example.js
```diff
@@ -5,7 +5,7 @@ var Diamond = function () {
     for (var i = 0; i <= inputIndex; i++) {
       output += getLine(inputIndex, i);
     }
-    for (var i = inputIndex - 1; i >= 0; i--) {
+    for (i = inputIndex - 1; i >= 0; i--) {
```
rchavarria: In this case, `i` is used in two different loops. I prefer not using the same variable to control two or more different loops, just in case I don't pay attention and I forget to initialize it in the second loop.

Instead of removing the second declaration, what about using a different name in the second loop? would it make the solution more readable?

exercises/sum-of-multiples/sum-of-multiples.spec.js
```diff
@@ -1,39 +1,39 @@
-var SumOfMultiples = require('./sum-of-multiples');
+var sumOfMultiples = require('./sum-of-multiples');
```
rchavarria: I would prefer having the variable starting with capital `S`. ESLint is complaining because the variable is used as a regular function, but in fact `SumOfMultiples` is a constructor function.

JavaScript allows to use constructor functions in both ways `SumOfMultiples()` and `new SumOfMultiples()`.

In this case, as `SumOfMultiples` is a constructor function, in order to fix ESLint errors, I would write `new SumOfMultiples([3, 5])` in line 5 for example. Does it makes sense to you?

If `SumOfMultiples` were a regular function, it would make a lot of sense to start it in lowercase.

tejasbubane:
@saramartinez We have now added eslint check in the CI and an [`eslintignore` file](https://github.com/exercism/javascript/blob/master/.eslintignore) with list of unimplemented exercises. We now expect contributors fixing the lint errors, to remove that particular exercise (`diamond` and `sum-of-multiples` in your case). Would you mind [rebasing your branch](https://www.atlassian.com/git/tutorials/rewriting-history/git-rebase) on master and removing them from the eslintignore?

Please feel free to let us know if this sounds unfamiliar or you are stuck with some problem.

saramartinez:
Thanks, I made the requested changes.

rchavarria:
Thanks a lot!! Great work!

PR #458, opened 2017-11-20
fix the trinary exercise

Radix16:
#399

### DIFF
--- a/exercises/trinary/example.js
+++ b/exercises/trinary/example.js
@@ -14,5 +14,5 @@ Trinary.prototype.toDecimal = function () {
 };
 
 Trinary.prototype.accumulator = function (decimal, digit, index) {
-  return decimal += digit * Math.pow(BASE, index);
+  return decimal + digit * Math.pow(BASE, index);
 };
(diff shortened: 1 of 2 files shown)


exercises/trinary/example.js
```diff
@@ -14,5 +14,7 @@ Trinary.prototype.toDecimal = function () {
 };
 
 Trinary.prototype.accumulator = function (decimal, digit, index) {
-  return decimal += digit * Math.pow(BASE, index);
```
rchavarria: The error ESLint throws is

```
/home/txingo/repositories/exercism-repositories/javascript/exercises/trinary/example.js
  17:3   error  Return statement should not contain assignment  no-return-assign
  17:10  error  Assignment to function parameter 'decimal'      no-param-reassign
```

It might be due to `+=`. And it makes sense. What's the point of assignning a value and then return it? Why not return it directly?

```
return decimal + digit ...
```

I tested, and just removing the `=` sign makes ESLint happy, and tests still pass.

rchavarria:
Sorry, I just forgot about this PR. It seems ok for me, so I'm gonna merge it.

Lots of thanks @Radix16!! Sorry for the delay.

PR #409, opened 2017-09-12
Port connect exercise from ECMAScript track

AdhirRamjiawan:
@rchavarria @matthewmorgan Implemented the following to resolve issue #389. Please let me know if all is well.

PR #461, opened 2017-11-21
Fix issues with connect PR

matthewmorgan:
This PR is to resolve a few issues with @AdhirRamjiawan 's pull request #409 to add the connect exercise.

I'm closing that PR in favor of this one.

A few things that I fixed:

1. Conflicts in config.json
2. An error in the `neighbors` method
3. Moved the class methods from being declared on an instance variable to being declared on the prototype.
4. Fixed some lint issues (which may have been mine.)
~~5. Added `package-lock.json` file.~~ removed as not needed

@exercism/ecmascripts please let me know if you have any concerns, otherwise I'm going to merge this PR in the morning and put it to bed.  Thanks!  @AdhirRamjiawan thanks for your work on this one.


package-lock.json
```diff
@@ -28,9 +28,9 @@
       }
     },
     "ajv": {
-      "version": "5.3.0",
```
rchavarria: I'm not a big fan of committing this kind of file, but I don't know how `package-lock.json` file works exactly, so I'm ok with this.

These changes could be done in a different PR, but I don't want to delay more the PR because of this personal preference.

So, ok, no change required.

exercises/connect/example.js
```diff
@@ -0,0 +1,84 @@
...
+  for (var i = 0; i < positions.length; i++) {
+    if (this.search(positions[i], player, [])) {
+      return true;
+    }
+  }
+  return false;
+};
+
+Board.prototype.search = function (pos, XorO, isChecked) {
+  var boardInstance = this;
```
rchavarria: I like this name. In other scenarios, I've seen it shorter: like `self` or `that`. What do you think of those names? They could be kind of a convention.

exercises/connect/example.js
```diff
@@ -0,0 +1,84 @@
...
+  }
+  if (this.winningSpot(pos, XorO)) {
+    return true;
+  }
+  var checked = isChecked.slice(0);
+  checked.push(pos);
+
+  var matches = this.neighbors(pos).filter(function (cell) {
+    return boardInstance.matches(cell, XorO) &&
+      checked.filter(function (spot) { return spot.x === cell.x && spot.y === cell.y; }).length === 0;
```
rchavarria: What about chainning a couple of filters?

```javascript
checked
    .filter(function (spot) { return spot.x === cell.x; })
    .filter(function (spot) { return spot.y === cell.y; })
    .length === 0;
```

Does it the code clearer?

exercises/connect/example.js
```diff
@@ -0,0 +1,84 @@
...
+  });
+
+  if (matches.length === 0) {
+    return false;
+  }
+
+  return matches.filter(function (spot) {return boardInstance.search(spot, XorO, checked);}).length > 0;
```
rchavarria: I feel this line is a bit long.

An alternative could be to extract a local variable, but I'm not sure it would make the code clearer. What's your opinion?

package-lock.json
matthewmorgan: I committed this after my Travis build failed with dependency problems.  I'm not sure if it's necessary or not-- it might be that my `node_modules` contained something out of date.  

I noticed that the Node version is Travis seemed to be unpinned and my CI build ran under NodeJS 9.2.  I'm wondering if we've made a mistake in not being specific about that version to prevent mysterious issues with the builds.   What do you think?

exercises/connect/example.js
matthewmorgan: I agree that it's a bit long, and we also have a local `matches` and a method `matches`.  I think I should change that.

exercises/connect/example.js
matthewmorgan: I agree that it reads better.  It also means every element potentially gets evaluated twice.  I don't think this is a big deal in this exercise, but maybe as a best practice it would be better to extract both check into one named function.

exercises/connect/example.js
matthewmorgan: This is only needed inside some functions that are nested, and thus bring their own `this`.  I tried to get rid of it entirely, but couldn't figure out how in ES5!

Let's try `self`.  I like that.

rchavarria: Yeah. Fat arrow functions solve most of these problems.

exercises/connect/example.js
rchavarria: It sounds to me like *premature optimization* 😝 

I usually favor readibility over performance (when performance is not an issue).

Even in real projects, if you're looping 100 items, it's almost the same looping them once than twice. Anyway, a well named function is much better than lot of things 😄

package-lock.json
rchavarria: I thought someone on @exercism/ecmascripts configured Node version to use the latest LTS (I'm not blaming, so I don't want to look for who was exactly). So, Node.js v8 should be used in Travis CI.

We can have `package-lock.json` in this PR, don't waste time with this.

matthewmorgan: Thanks for having a look Rubén!

rchavarria:
Merging!!

PR #463, opened 2017-12-02
alphametics: resolve eslint errors

Smarticles101:
ref #339

### DIFF
--- a/exercises/alphametics/example.js
+++ b/exercises/alphametics/example.js
@@ -53,21 +53,21 @@ function testNumbers(numbers, puzzleParts, firstLetters) {
   }
   var replaceRegex = new RegExp('[' + keys.join('') + ']', 'g');
 
-  puzzleParts = puzzleParts.join(',')
+  var puzzlePartsJoined = puzzleParts.join(',')
     .replace(replaceRegex, function (input) { return numbers[input]; })
     .split(',')
-    .map(function (t) {return parseInt(t);});
+    .map(function (t) {return parseInt(t, 10);});
 
-  var total = puzzleParts.slice(puzzleParts.length - 1)[0];
-  return total === puzzleParts
-    .slice(0, puzzleParts.length - 1)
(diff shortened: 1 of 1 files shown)


exercises/alphametics/example.js
```diff
@@ -53,21 +53,21 @@ function testNumbers(numbers, puzzleParts, firstLetters) {
...
+  return total === puzzlePartsJoined
+    .slice(0, puzzlePartsJoined.length - 1)
     .reduce(function (acc, val) { return acc + val; }, 0);
 }
 
 function getPermutations(inputArr) {
   var results = [];
-  function permute(arr, memo) {
-    var cur, memo = memo || [];
+  function permute(arr, memo = []) {
```
matthewmorgan: [Default parameters](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Default_parameters) are an ES6 feature. , so for the time being we want to avoid using them in this track.

How about:

```javascript
function permute(arr, memo) {
  memo = memo || [];
  var cur = memo;
  ...
}
```

Smarticles101: Sounds good, however what you suggested leaves a `no-param-reassign` eslint error

matthewmorgan: Yes, I see.  I think the right approach is to change the initial method call to pass in an empty array for `memo`, like so:

```javascript
function getPermutations(inputArr) {
  var results = [];
  function permute(arr, memo) {
    var cur = memo;
    for (var i = 0; i < arr.length; i++) {
      cur = arr.splice(i, 1);
      if (arr.length === 0) {
        results.push(memo.concat(cur));
      }
      permute(arr.slice(), memo.concat(cur));
      arr.splice(i, 0, cur[0]);
    }
    return results;
  }
  return permute(inputArr, []);
}
```

Smarticles101: Yep that works great, I just pushed a commit with that change :)

matthewmorgan:
Thanks @Smarticles101 !!

PR #467, opened 2017-12-08
Fix ESLint errors for `circular-buffer` exercise

rchavarria:


exercises/circular-buffer/circular-buffer.spec.js
```diff
@@ -57,7 +57,7 @@ describe('CircularBuffer', function () {
   xit('writes of undefined or null don\'t occupy buffer', function () {
     var buffer = circularBuffer(3);
     buffer.write(null);
-    buffer.write(undefined);
+    buffer.write(undefined); // eslint-disable-line no-undefined
```
rchavarria: I disabled ESLint for just this line. ESLint doesn't like undefined, but in this test, undefined is a value we want to test, so deleting the line is not an option.

matthewmorgan:
Thanks @rchavarria !

PR #469, opened 2017-12-08
Fix ESLint errors for `food-chain` exercise

rchavarria:


rchavarria:
I liked more how `food-chain/example.js` looked before. But ESLint requires `switch` statements to have a `default` case and every function to finish with a `return` statement.

I'm open to hear about other alternatives.

matthewmorgan:
Looks fine to me.  I would be OK to merge this as it currently is.

exercises/food-chain/example.js
```diff
@@ -29,51 +29,64 @@ FoodChain.prototype.verses = function (first, last) {
...
+            'I don\'t know why she swallowed the fly. Perhaps she\'ll die.\n';
+    break;
+
+  case 8: verse = 'I know an old lady who swallowed a horse.\n' + 'She\'s dead, of course!\n';
+    break;
+
+  default: verse = '';
   }
+
+  return verse;
```
tejasbubane: I think we can remove the variable `verse` and return from each case. That would remove the multiple `break`s.

matthewmorgan: I used to dislike having multiple `returns` in a function.  I have changed my mind about that though, especially in a structure like this where the semantics are really clear.

Also interesting to note is that I've run some cyclomatic complexity analyzers on code like this, and as far as they're concerned, ` if x 10` with one `return` is the same complexity as having 10 `return`s in one function.

rchavarria: The previous version had as many returns as `case`s.

I was happy with that version (I don't remember who was the author), but ESLint didn't think like that. ESLint *forced* me to add a `default` case and a return value. That's why I created a local variable.

What I would discourage is to mix them: I want all `case`s (even the `default`) equal. All or none shoudl return value. To fix ESLint errors, it looked betteer to me to return just at the end of the method.

matthewmorgan: Hmm.  I think this is a case where I don't exactly agree with the linting rules, but I don't feel strongly enough about that to suggest we make an exception for this exercise.

I am back to my original position: merge as is!

rchavarria: I feel exactly like that. I don't like the rules in this case, but they are the rules we chose. So, it's some code I need to bear with. I'm not the code, so I need to understand that sometimes, it's good not to agree completely on how the code looks like if it's better for the team and for the long run.

PR #471, opened 2017-12-10
fix anagram exercise

Radix16:
#399


tejasbubane:
Thanks @Radix16. Looks good. 👍

PR #472, opened 2017-12-10
fix food chain exercise

Radix16:
#399


Radix16:
PR is already pending for this work.  I should pay more attention.

rchavarria:
😆 😆 

It happens from time to time. Don't let that to stop you 💪

Issue #488, opened 2018-01-07
bowling exercise does not match the canonical tests


danielj-jordan:
The current example and tests for the bowling exercise do not canonical [specs](https://github.com/exercism/problem-specifications/blob/master/exercises/bowling/canonical-data.json) or the trusted version of the [readme](https://github.com/exercism/problem-specifications/blob/master/exercises/bowling/description.md).

The readme specifically states that function roll(pins : int) should exist. It does not exist in the sample.  Also there are a few tests that are not implemented in the spec.js file which are listed in the canonical specs. One of which is listed here:
"description": "bonus roll after a strike in the last frame cannot score more than 10 points", "property": "roll", "previous_rolls": [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 10], "roll": 11, "expected": {"error": "Pin count exceeds pins on the lane"}
This example and spec should be re-written to match the canonical version.

PR #496, opened 2018-01-21
redo the bowling exericse to match the canonical tests and description

danielj-jordan:
#488 
the bowling exercise now strictly matches the canonical [tests](https://github.com/exercism/problem-specifications/blob/master/exercises/bowling/canonical-data.json)


matthewmorgan:
@danielj-jordan sorry to see this has lingered so long.

It looks as if in the meantime two more tests have been added to the canonical set, but let's leave that for another PR.

Thanks for the great work!

Issue #462, opened 2017-11-25
Use LTS node on this track


matthewmorgan:
I thought we had decided:

1. `lts/*` for JavaScript track
2. `node` for EcmaScript track

Please let me know.  If this is the case, we need PRs in both tracks to fix the `.travis.yml` files, which currently are `7.8` for ES, and `node` for JS.

@exercism/ecmascripts

joelwallis:
Makes sense, but since the EcmaScript track uses Babel to compile classes and outputs code compatible with older versions. If we are looking to see if the current Node.js version is able to execute the code this issue should be addressed.

rchavarria:
It makes completely sense for me too. `lts/*` for JS, `node` for ES.

matthewmorgan:
@joelwallis I see the merit in your thought, but as a maintainer I can't imagine trying to keep everything current with latest stable Node without transpilation.

The users never _see_ the transpiled code, since it's all done magically by Jest, so I'm not sure there's much value in trying to differentiate what will work or not work with a given version of Node. 
 
But, to make sure everything works as expected on this track: it looks like a preset block like this would ensure our Node "latest" gets what it needs from Babel:

```json
{
  "presets": [
    ["env", {
      "targets": {
        "node": "current"
      }
    }]
  ]
}
```

I vote for changing node to "latest" in Travis, and adding this preset for Babel.  I believe this will also correctly target the user's installed Node version if it happens to not be the same as ours.

rchavarria:
Go for it!

tejasbubane:
I guess this can be closed now since https://github.com/exercism/ecmascript/pull/390 is merged.

PR #497, opened 2018-02-01
Use LTS version of node in travis-ci

tejasbubane:
As suggested in https://github.com/exercism/javascript/issues/462.


tejasbubane:
I am going ahead to merge this since maintainers seem to agree on this in #462

PR #499, opened 2018-02-12
Fixes: Alphametic Lint Fixes

ajomadlabs:
Tested both alphametics.spec.js and example.js .
Found no eslint errors in both files.
Removed these files from .eslintignore.


matthewmorgan:
Thanks @ajomadlabs !  We appreciate your contribution.

PR #502, opened 2018-02-14
Eslint configuration

### LINKED ISSUE #456: Move ESLint configuration to `package.json` file
As discussed on #441, let's move ESLint configuration to `package.json` file.

In summary, [ecmascript](https://github.com/exercism/ecmascript) track has already ESLint configuration in `package.json` files, because it was more convenient for track maintenance.

In the case of this track, it's not so important where the configuration is placed. But, just to keep consistency between those two tracks, ESLint configuration should be inside `package.json` file.

ajomadlabs:
Fixes #456

### DIFF
--- a/nano.save
+++ b/nano.save
@@ -1,0 +1,6 @@
+Fixes: Binary-search-tree Lint Fixes
+
+Tested the files, found an error
+Fixed it by changing 'undefined' to 'null'
+Removed the dile from .eslintignore
+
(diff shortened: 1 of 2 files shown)


nano.save
```diff
@@ -0,0 +1,6 @@
+Fixes: Binary-search-tree Lint Fixes
```
matthewmorgan: Hmm... we don't want this file in the repo please!

ajomadlabs:
@matthewmorgan sure

tejasbubane:
@ajomadlabs You can do the fixes and push on this same branch so that the changes appear here. Saves creating a new PR and keeps all the communications in one place.

Issue #505, opened 2018-02-27
THis is more of a question about Jasmine??


AnnahGerletti:
I am new to using Jasmine, I find when something breaks in the code Jasmine does not always out-put an error to my terminal. Why is this. I just get the same line returned.

like this:
```
➜  rna-transcription git:(master) jasmine rna-transcription.spec.js
➜  rna-transcription git:(master) jasmine rna-transcription.spec.js
➜  rna-transcription git:(master) jasmine rna-transcription.spec.js
➜  rna-transcription git:(master)
```
The first line is good with a green arrow, but the next three are the responses to running the command.
Nothing happens. No Output.

joelwallis:
Question previously asked on exercism/exercism.io#3732

joelwallis:
Hi @AnnahGerletti,

I'm one of the co-maintainers of the JavaScript track. This Jasmine behavior definitely isn't normal. It's supposed to show something like this:

```
$ jasmine ./hello-world.spec.js
Started
.


1 spec, 0 failures
Finished in 0.009 seconds
```

Which operating system are you using? Also, can you provide more information about your environment? Run the command below and paste the output here:

```
exercism version; node --version; jasmine -v
```

AnnahGerletti:
Hello, 

I am running on a MacOS, using Zsh as my CLI, writing code in VSCode. Is that what you wanted to know. 

here is the results to the comand
```
➜  rna-transcription git:(master) exercism version; node --version; jasmine -v
No help topic for 'version'
v9.4.0
jasmine v3.0.0
jasmine-core v3.0.0
➜  rna-transcription git:(master) 
```

AnnahGerletti:
There will be times when I do see something like this:
```
Randomized with seed 53901
Started
....F**

Failures:
1) toRna() transcribes all dna nucleotides to their rna complements
  Message:
    Expected undefined to equal 'UGCACCAGAAUU'.
  Stack:
    Error: Expected undefined to equal 'UGCACCAGAAUU'.
        at <Jasmine>
        at UserContext.<anonymous> (/Users/annahgerletti/exercism/javascript/rna-transcription/rna-transcription.spec.js:23:8)
        at <Jasmine>
Pending:
```
However if I change the code in VScode and then try to run the test again, I won't get any kind of error message which is difficult to debug. I was just wondering are there steps when using Jasmine that I need to keep in mind? Forexple can't run partial code, or no comment? things like that.

joelwallis:
[...]

PR #507, opened 2018-03-01
Update jasmine to v3.1.0

tejasbubane:
Recently [an issue](https://github.com/exercism/javascript/issues/505) came up related to jasmine v3.0.0. And I noticed that we are using v2.8. Jasmine 3 seems to have some [breaking changes](https://github.com/jasmine/jasmine/blob/master/release_notes/3.0.md) and since many of our new users will be installing recent version of jasmine, I thought it would be better to update it so we run CI using v3.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -728,12 +728,6 @@
       "integrity": "sha1-Cr9PHKpbyx96nYrMbepPqqBLrJs=",
       "dev": true
     },
-    "exit": {
-      "version": "0.1.2",
-      "resolved": "https://registry.npmjs.org/exit/-/exit-0.1.2.tgz",
-      "integrity": "sha1-BjJjj42HfMghB9MKD/8aF8uhzQw=",
-      "dev": true
-    },
     "external-editor": {
       "version": "2.1.0",
       "resolved": "https://registry.npmjs.org/external-editor/-/external-editor-2.1.0.tgz",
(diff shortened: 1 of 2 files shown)


tejasbubane:
A further step would be to test against multiple versions of jasmine in the CI and make sure our tests pass with all major versions of jasmine - although I don't think this is required and I even don't know how to do this - I will wait for opinion from other people on this.

Also that particular issue was related to jasmine itself and could not have been caught even if we tested against multiple versions.

matthewmorgan:
@tejasbubane this change makes sense to me.  I'm glad to see that it didn't create any issues with the build.

I'm happy to merge this, but let's see if @rchavarria or @joelwallis have any concerns or ideas.

joelwallis:
Hm, just found something that could be a potential issue. I tried to run our tests from my Moto G4 Plus (no reason besides fun on seeing my phone burning during the tests) and I couldn't make it to work at first time. Reason? Unable to find `jasmine`.

What I believe to be the reason is that `make` don't properly setup npm's local binaries to be included in `PATH`. So, make we could execute a different command from `make` to get local `jasmine`, something like:

```
PATH="$PATH:`pwd`/node_modules/.bin" jasmine --random=true $(OUTDIR)/$(TSTFILE)
```

I don't think it's something that should be blocking this PR. Instead, I could open a new one to tackle it.

@matthewmorgan @tejasbubane thoughts?

joelwallis:
By the way, my Moto G4 Plus just ran the tests and I got an interesting error (screenshot below). Also, it doesn't block the pull request because tests are passing on my MacBook Pro but it makes think about how ARM architectures deal with complex numbers 😄

screenshot_20180304-122811

PS: I ran it twice and got the same error.

matthewmorgan:
@joelwallis I think a separate PR would be great.

As far as the rounding discrepancy, we could use a smarter matcher.  Here is the one I think we would want: https://github.com/jasmine/jasmine/blob/master/src/core/matchers/toBeCloseTo.js

What do you guys think?

joelwallis:
Wow, I didn't know there was an assertion method like this one. That would be awesome to have it instead of a precision comparison! How many decimal points would make sense for us?

matthewmorgan:
@joelwallis I don't know, but the disagreement in the above is way out at 16-17 decimal places.

12?

tejasbubane:
@joelwallis That precision bug is a great find! 😄 👍 I suggest we revisit the all the test cases for that exercise and use `closeTo` wherever appropriate.

Although I am a little unsure about the path issue since we recommend users to install jasmine globally for this track - not sure if we should switch to local `(node_modules/.bin)` installation.

tejasbubane:
I think both Mathew and Joel are okay with this PR - merging.

PR #517, opened 2018-03-01
Fix lint errors in robot simulator

ajzdunek:
changed line 1 variable to robotOne and module.exports added default this.coordinate to switch statement.

### DIFF
--- a/exercises/robot-simulator/example.js
+++ b/exercises/robot-simulator/example.js
@@ -1,4 +1,4 @@
-var Robot = (function () {
+var robotOne  = (function () {
   'use strict';
 
   var VALID_DIRECTIONS = ['north', 'east', 'south', 'west'];
@@ -39,6 +39,8 @@ var Robot = (function () {
     case 'west':
       this.coordinates[0]--;
       break;
+    default:
+      this.coordinates[1]++;
     }
   };
 
@@ -84,4 +86,4 @@ var Robot = (function () {
   return Robot;
 })();
 
-module.exports = Robot;
+module.exports = robotOne;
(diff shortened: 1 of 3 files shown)


matthewmorgan:
@ajzdunek thank you for your contribution.

What issue does this PR attempt to address?  Why do you propose these changes? Can you please explain in the description?

exercises/robot-simulator/example.js
```diff
@@ -1,4 +1,4 @@
-var Robot = (function () {
+var robotOne  = (function () {
```
tejasbubane: Can you remove this function wrapping? This should remove the need for this extra variable - the name `robotOne` seems a little odd. Should look something like this:

```js
'use strict';

var VALID_DIRECTIONS = ['north', 'east', 'south', 'west'];
var INSTRUCTION_KEYS = {
  A: 'advance',
  L: 'turnLeft',
  R: 'turnRight'
};

function Robot() {
  this.coordinates = [0, 0];
  this.bearing = 'north';
}

Robot.prototype.at = function (x, y) {
  this.coordinates = [x, y];
};
....
....
```

If you are confused, here is how the entire file will look like: https://gist.github.com/tejasbubane/ef04a2d3d5f88858f6f2ddd1c8b6c92b plus the default condition change which you have in this PR.

tejasbubane:
@matthewmorgan This PR is for fixing lint errors for the `robot-simulator` exercise. There were two errors in the file:

```
  11:12  error  'Robot' is already declared in the upper scope  no-shadow
  29:5   error  Expected a default case                         default-case
```

The default case fix looks good. And the shadow-variable fix can be done better by removing the variable rather than changing the name.

matthewmorgan:
@tejasbubane I agree.  I thought there might be some value in getting @ajzdunek in on the conversation, but for the record, I would have suggested the same changes you did. 🥇

tejasbubane:
@ajzdunek Please feel free to reopen if you wish to continue working on this one.

Issue #531, opened 2018-03-21
Canonical test spec for bowling excercise has two new tests


danielj-jordan:
The [canonical spec](https://github.com/exercism/problem-specifications/blob/master/exercises/bowling/canonical-data.json) has two new tests.   Implement the tests for “cannot roll after bonus roll for spare” and “cannot roll after bonus rolls for strike” which have been added to version 1.2.0 of this document.

PR #532, opened 2018-03-22
implement the two recently addeded tests from the canonical test spec

danielj-jordan:
#531


matthewmorgan:
Lovely.  Thanks @danielj-jordan !!

PR #533, opened 2018-03-22
remove linted exercises from lint ignore

tarunvelli:
removed custom-set, flatten-array and queen-attack exercises from .eslintignore as they are properly linted. also currently #399 shows grade-school as still needing code fixes, but even that is already fixed. please update the list accordingly.


matthewmorgan:
Thanks @tarunvelli !!

PR #535, opened 2018-03-29
Fix robot simulator

leahzxxz:


exercises/robot-simulator/example.js
```diff
@@ -1,87 +1,85 @@
...
-  'use strict';
+'use strict';
 
-  var VALID_DIRECTIONS = ['north', 'east', 'south', 'west'];
-  var INSTRUCTION_KEYS = {
-    A: 'advance',
-    L: 'turnLeft',
-    R: 'turnRight'
-  };
+const VALID_DIRECTIONS = ['north', 'east', 'south', 'west'];
```
matthewmorgan: We're not using `const` or `let` in the JavaScript track yet.  At the moment we're reserving those features for the EcmaScript track, and this track should be ES5 only.  Can you please change all `const` and `let` to `var`?  Thanks!

leahzxxz: Hi! Sorry about that - I didn't know it was ES5 only. Will change everything back. Thanks!

matthewmorgan:
Great.  I think we could (should?) add linting rules to prevent the use of ES6 features.  Thank you for the PR!

PR #541, opened 2018-04-24
Use npm-ci instead of npm-install

tejasbubane:
Looks like `npm-ci` is meant to be used in CI environments.
https://docs.npmjs.com/cli/ci

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ node_js:
   - "lts/*"
 
 install:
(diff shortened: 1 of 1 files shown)


tejasbubane:
Ok `npm ci` command was [added in npm v5.7.0](https://github.com/npm/npm/releases/tag/v5.7.0). But we are using `5.6.0` along with node LTS.

PR #543, opened 2018-04-27
simplify instructions and module structure

alexch:
I taught a small class this week using these exercises and the two
students had a lot of trouble following along with the instructions. As
a proof of concept, I revised two lessons with the following
improvements:

  * move any boilerplate or comments into the README, allowing
    students to create an empty source file instead of struggling
    with a partially-filled-in stub source file

  * remove redundant setup instructions, which clutter the file and
    make it harder for students to find their place again in the
    instructions and hints for the problem at hand -- setup instructions
    are available at the exercism.io web site and do not need to appear
    inside every lesson

  * use a minimal module declaration (`module.exports = function()`)
    -- when the exercise demands a function, the solution should
    be a function, not an arcane nested object structure that may or
    may not contain a function (whose name is confusingly similar to
    the name of the parent object)

  * remove `xit` and instead advise using `jasmine --fail-fast=true`
    which accomplishes the goal of focusing attention on a single
    failing test at a time without requiring the student to understand
    Jasmine syntax, how to correctly search-and-replace, and to edit the
    otherwise read-only spec file

What do you think of these suggestions? If I teach another class with
these, I will make similar changes to the rest of the exercises and add
them to this branch, but I thought I'd start by showing you what I've
done and asking for your feedback in advance of any more thorough PRs.

I know you've discussed and debated a lot of these issues before and I
don't want to seem pushy or holier-than-thou. I'm coming to this with my
own experience as a teacher and test-driven-lesson creator (I created
testfirst.org with Sarah Allen and have taught many classes using
test-guided exercises in Ruby and Java as well as JS) and with my own,
possibly peculiar needs to teach in-person test-first classes.


gavinhenderson:
> use a minimal module declaration (module.exports = function())
-- when the exercise demands a function, the solution should
be a function, not an arcane nested object structure that may or
may not contain a function (whose name is confusingly similar to
the name of the parent object)

Personally, I think this is worth making into a seperate issue. This is something that annoys me quite a lot and if we can get a consensus on using your approach we will be able to push forward and covert the whole course to it.

alexch:
> When you say 'when the exercise demands a function...' can you explain that more for me? Are you talking about helloWorld.hello() for example? That is potentially confusing.

Yes, exactly. In the context of a lesson on functions, the solution should be a function -- not a method on the prototype of a class that (because JavaScript) is *also* a function but not *the* function. Whereas `module.exports =` is adding one extra (but necessary (because JavaScript)) concept, but it's much easier to understand or at least memorize or ignore than the current export incantation.

> Also, When you say 'redundant instructions,' do you mean the boilerplate that appears in all exercises, or were you referring to something else?

Yes, that's what I mean. The README setup instructions are redundant with both the web site and with each other. As with boilerplate stub/template source code, it makes the student's cognitive load heavier -- I have to remember where I was inside a longer document, and also realize that when I find these instructions I should actually *not* follow them, because I've already installed Node/Jasmine/etc. 

> if we can get a consensus on using your approach we will be able to push forward and covert the whole course to it

That is my intention with this PR. The conversion should be fairly easy and mechanical; I'm happy to do it if we agree on the concept(s).

Thanks for taking the time to review and ponder :-)

matthewmorgan:
Closing as wontfix per #556.

